PR #112, opened 2022-06-08
Fixed neovim hover issue

JamesDSource:
These changes fix an issue with neovim that caused the editor to flash a big annoying error whenever you tried to hover over something invalid. Also conforms with the lsp specification slightly better.


DanielGavin:
Thanks. Yeah think it's neovim lsp only.

PR #128, opened 2022-08-01
Implement platform independent `run_executable`

igordreher:
This should make diagnostics work on other platforms besides Windows.


DanielGavin:
Thanks.

PR #129, opened 2022-08-12
fixed memory corruption with checker_args

IanLilleyT:
I was getting random crashes and it turns out the `checker_args` was using garbage memory, effectively fuzzing the odin calls


DanielGavin:
Thanks!

PR #131, opened 2022-08-12
translate logger level to diagnostic level

IanLilleyT:
the client wasn't showing the correct log levels because it was always passing the error level

but with this change, the `Starting Odin Language Server` message won't show up because it's info, not error, so maybe the min log level should be info, like:
```
context.logger = server.create_lsp_logger(&writer, log.Level.Info)
```

the only other log.info is:
```
log.infof("document_close: %v", uri_string)
```
i don't know how much noise that causes, but if it's noisy it could be made debug


DanielGavin:
Might be fine. I was worried that it it could cause performance issues, but let's see.

PR #151, opened 2022-09-18
[vscode] Fix debugging in vscode ide

flplv:
[VSCODE only]
This CR fixes the error when we click the "Debug" decoration button on top of a `@(test)` decorated function;
it also activates vscode's support to add breakpoint to the language

### DIFF
--- a/editors/vscode/package-lock.json
+++ b/editors/vscode/package-lock.json
@@ -1,8 +1,3860 @@
 {
 	"name": "ols",
 	"version": "0.1.2",
-	"lockfileVersion": 1,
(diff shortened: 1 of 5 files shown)


DanielGavin:
Sweet thanks. It seems the Ci is crashing on tests that has nothing to do with your changes. Not sure if this is something introduced with a new version of Odin.

DanielGavin:
I'll make a new version of the extension.

flplv:
wow that was fast...

flplv:
I got one more commit that may be controversial.. I will open another PR now

PR #152, opened 2022-09-18
[vscode] icon fix; plus small change

flplv:
and 
[minor] change the promise handling style of fs.stat


flplv:
btw: feel free to reject, in case you want to maintain the ols.json location where it is.

I figured every extension in vscode uses the .vscode folder for settings so we'd like to do that too

flplv:
Just pushed another patch fixing the file icon.
I chose the de-saturated blue from the odin website as color because it works ok for both light and dark themes

DanielGavin:
We can't move the path of the `ols.json`, it is used to indicate the root of the project.

flplv:
here it is a screenshot of how vscode looks now: 
Screenshot from 2022-09-19 16-47-40

flplv:
another one with the debugging working :)
Screenshot from 2022-09-19 16-52-31

PR #155, opened 2022-10-06
Add snippets for VSCode

rasa-silva:
I think something like this should be useful...
Consider this as a suggestion more than a finalised PR 😁


DanielGavin:
You should try to use the snippets through the LSP.

src/server/snippets.odin

That way all the editors can use them.

rasa-silva:
Oh, didn't know that LSP supported snippets but that makes sense...I'll close this and open another one with more snippets 👍🏻

PR #176, opened 2023-03-04
Update odinfmt with new filepath.Walk_Proc signature

hwchen:
Per https://github.com/odin-lang/Odin/commit/9d50a0490530f8c602ece4c9bd56f316e169b66a

an additional parameter `user_data: rawptr` was added to the `filepath.Walk_Proc` callback.

This commit updates odinfmt to use the new parameter.


DanielGavin:
Thanks!

PR #178, opened 2023-03-10
Improve windows update

fabiansperber:
Put ols binaries into their own folder by release id, so an update can be done in parallel while still being in use (windows)


DanielGavin:
Thanks. I'll build it later today.

PR #183, opened 2023-04-08
vs-code: add darwin arm support

limaak:
vs-code: add darwin arm support

Done:
- Add arm to supported platforms => https://github.com/DanielGavin/ols/pull/183/files#diff-16f6ad44a1c2080b7fffb341305a0968e2944905b1a8211044d331d2abf63ebdR263.
- Formatting

Tested on Apple M1 Pro

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # ols
-Language server for Odin. This project is still in early development. 
+Language server for Odin. This project is still in early development.
 
 **Status**: All platforms work.
 
@@ -18,10 +18,10 @@ Language server for Odin. This project is still in early development.
  ```
  cd ols
 
- //for windows
- ./build.bat 
+ # for windows
+ ./build.bat
 
- //for linux
+ # for linux/mac
  ./build.sh
  ```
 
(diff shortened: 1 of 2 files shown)


DanielGavin:
Arm is already supported by the vscode extension, you just need manual set the `ols` binary path. If you want to support it with auto downloading the binary, you need to also figure out how to build the arm binary in the night CI.

limaak:
> Arm is already supported by the vscode extension, you just need to manually set the `ols` binary path. If you want to support it with auto downloading the binary, you need to also figure out how to build the arm binary in the CI.

Screenshot 2023-04-09 at 19 47 47

By adding following changes, the error is gone, however we will still need to setup `ols` binary path.

DanielGavin:
That error should go away if you follow the message. I don't want that message to go away, I need to inform the user that they have not set the `ols.server.path`. When they have done that, it should give away, and if it's not, then that is an error. 

The code you modified is directly related to the shipped binary logic, and that should not be touched until we have a shipped binaries for arm darwin.

limaak:
@DanielGavin thanks for answer, I've closed the issue for now :)

DanielGavin:
@limaak Alright.

If you are feeling daring you could try to figure out how the build the arm binary to the nightly releases. 

It seems like they have released a CI runner for m1/m2: https://github.blog/changelog/2022-08-09-github-actions-self-hosted-runners-now-support-apple-m1-hardware/

It might also be possible to cross compile with `odin`, but I haven't tried that. That would be much more optimal, since you could then just build all the binaries from one runner.

PR #196, opened 2023-06-11
Add json-schema files for json config files

thetarnav:
I've included all fields present in the structs in the source, but maybe they should be limited to those mentioned in the README.

I have no idea if the schema is correct by any standard, but the autocomplete works well:

https://github.com/DanielGavin/ols/assets/24491503/2a0b27ea-ae01-4850-ba6c-1766929b1ecc

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = tab
+indent_size = 4
+end_of_line = lf
+charset = utf-8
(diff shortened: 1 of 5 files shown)


DanielGavin:
Great work.

The schema for config is looking at the wrong struct. It's: https://github.com/DanielGavin/ols/blob/4965a0c31082f84ce272a43548933d323e6427e4/src/server/types.odin#LL332C3-L332C3

PR #205, opened 2023-06-21
Improve textmate grammars for procedure declarations

thetarnav:
Took some time trying to improve the textmate grammars for procedure declarations.

image

It's not perfect, I'm still not sure how to do these parts:
- differentiate between procedure types and normal procedure declaration (right now they are both functions)
- supporting assigning with type specified (`my_proc: Type : proc() {}`)
- differentiating between named return types and unnamed return types

Some of the patterns here could probably be reused for variable/struct declarations as well


DanielGavin:
Thanks. I have no experience with textmate grammars, so it's better than what I could do.

PR #207, opened 2023-06-23
Add more builtin procedures

thetarnav:
Adds some of the missing procedures from the `builtin` package.
Also it moves the ones that were in `keyword.function.odin` to `support.function.builtin.odin` as I don't see why they should have a different scope — they come from the same package.


thetarnav:
I'm not sure if we should also include procedures from the `runtime` package to the list here or rely on semantic tokens for that. Some of those methods are already included *(`make|resize|reserve|append|delete|assert|panic`)*, but that's only a very small section of all runtime procedures.
All these procedures are not really keywords either, they can be overwritten by the user, but the theming will remain, unless the semantic tokens help.

PR #208, opened 2023-06-24
Improve build time for debug builds

ryuukk:
From 4sec down to 1.6sec


DanielGavin:
Could be interesting to use the experimental features. It should be kept to the debug builds, I don't want to risk the release builds.

PR #215, opened 2023-06-29
Format labelDetails for procs that returns and variables/constants

ryuukk:
It gives a cleaner and less noisy look overall

Some screenshots:

image

image

image

image

image


DanielGavin:
looks great.

DanielGavin:
@ryuukk 

Maybe we should remove the `->`?

image

Not sure how I feel about it.

PR #223, opened 2023-07-06
Allow a little more line break control in composite literals

flga:
A composite literal whose declaration spans multiple lines will now be formatted as multi line, otherwise it's formatted on a single line.

This is a breaking change w.r.t. composite literals that do not have fields. Previously they would always be inlined, now they will only be inlined if they are already inlined.

A further improvement would be to format any composite literal while adhering to the user's line breaks, while still doing the whole wrapping business. I couldn't get this to work.

While not as close as I'd like, overall this change brings odinfmt's output more in-line with `demo.odin`, with user defined matrix and vector types being the odd man out.

Bonus: Fixed Allman style braces in composite literals, I don't know who would do this to themselves but now they can!


DanielGavin:
`
A further improvement would be to format any composite literal while adhering to the user's line breaks, while still doing the whole wrapping business. I couldn't get this to work.
`
This formatter should in generally never adhere to the users specific line break and goes against the formatters design. I mentioned on Discord to have an experimental config to break down call expressions when they were already broken in text. It was something I was skeptical of, but would be interesting to see through a config.

`
While not as close as I'd like, overall this change brings odinfmt's output more in-line with demo.odin, with user defined matrix and vector types being the odd man out.
`
Sounds like you want an equivalent of `gofmt` that does no formatting but whitespace changes, i.e. no linebreaking of any kind. 

I posted an old version of the formatter that looks a lot like `gofmt`. That is probably a better starting point then trying to hammer in these features you want. I can't see this ending well. Nothing wrong with having two formatters that are completely different.

flga:
Hi Gavin,

What we discussed on Discord was a much more ambitious vision, we reached the conclusion it was worth a shot to see how it would pan out.
This is the result, a very cut down version of the original proposition, it doesn't affect call expressions (directly) it just allows a user to effectively toggle between multi line and single line composite literal declarations.

This is the go/no go point, I ask you to review the PR as is, I have no plans to add more features.
Obviously, even though it's an isolated change, you may still argue it opens a precedent you're not comfortable with. I understand that and am fine with whatever you decide.

DanielGavin:
I can't merge it without it being configurable. The default formatter version can't change the current behavior. 

This is supposed to be experimental and therefore something people actively want to turn on.

flga:
@DanielGavin I've added a new config option to toggle this on `exp_multiline_composite_literals`. It's off by default.

If and when the feature gets accepted we'll need to drop the `ext_` prefix. That means we'll need to figure out what to do with the (by then) old option, probably just allow it and wire it up internally to cause minimal breakage.

Since we don't yet have the ability to configure the tests in a piecemeal fashion for now they just turn it on. Let me know if you'd like to follow another approach.

DanielGavin:
I guess I forgot to approve the review and the comment didn't show.

PR #231, opened 2023-07-24
Hack for sublime text bug

ryuukk:
Sublime text has an issue where it doesn't show completions for items that has both ``..`` and ``#`` in the label detail

https://github.com/sublimehq/sublime_text/issues/6033

This PR is a workaround, it basically replace ``..`` with a visual equivalent ``ꓸꓸ`` https://www.compart.com/en/unicode/U+A4F8

This should be removed once the issue gets fixed in Sublime Text

It hides lot of procs, specially in ``core:fmt`` module, so i think it worth having this hack for now, what do you think?

If not mergeable, I think it worth letting this open, at least as documentation for people who use Sublime Text

Easy to reproduce:

```odin
reproduce_the_bug :: proc(fmt: string, args: ..any, location := #caller_location) {
}
```

### DIFF
--- a/src/server/completion.odin
+++ b/src/server/completion.odin
@@ -1822,7 +1822,7 @@ format_to_label_details :: proc(list: ^CompletionList) {
 		// or if this PR gets merged: https://github.com/sublimelsp/LSP/pull/2293
 		dt:= &item.labelDetails.? or_else nil
 		if dt == nil do continue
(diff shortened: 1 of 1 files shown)


DanielGavin:
Yes. Try to find the client and only apply it to sublime. You can always save it in `common.config` global variable.

DanielGavin:
Had some conflicts. I'll merge it and add the editor check myself.

PR #238, opened 2023-08-09
update for switch _ in

hwchen:
With https://github.com/odin-lang/Odin/commit/cd74cdfdaf5157704b836c8de1c32b5a03dddcae, `switch in` is now an error, but ols was autoformatting `switch_ in` to `switch in`.

Fix autoformatting so `switch _ in` autoformats to `switch _ in`.


DanielGavin:
Great. Thanks!

PR #263, opened 2023-10-26
Remove writing semantic nodes for untyped and basic value symbols

thetarnav:
The result semantic token types were often incorrect for constants and literals like `nil`, `true` and `false`.
Syntax grammars have better results here, but were being overridden by the semantic tokens.

I'm not yet familiar enough with the codebase to fix this properly, instead of just removing it.


DanielGavin:
Think that's fine. I was planning on removing most of the semantic for things that really could be done with syntax grammar.

PR #269, opened 2023-11-11
respect //+private and the file/package variant

laytan:


laytan:
ci not my fault, made #270 to fix

PR #275, opened 2023-11-28
better bitset completion

laytan:
Adds more completion support for bitsets.

- "magic" completion when you type an expression that resolves to a bitset type and you use `.` it shows all possibilities in the bitset and when completion is confirmed removes the expression, leaving the `.Variant`. eg: `variable.` comes up with completions `variable.Foo`, `variable.Bar`, and confirming will remove the `variable`, leaving `.Foo`
- add bitsets to the value decl completion, `a : My_Bitset = { .* }` will show the bitset variants after typing the `.`
- add bitset variant completion when calling a procedure, eg: `hello({ . })` shows the variants after typing the `.`

For that last point I had to remove an early return which would return if there was a comp literal in the context. This royally screwed the diff though because I had to indent the whole block by one (because of surrounding it with an if). Nothing in that hunk has changed though, just turned the early return into a normal if statement.

### DIFF
--- a/src/server/completion.odin
+++ b/src/server/completion.odin
@@ -517,6 +517,40 @@ get_selector_completion :: proc(
 			append(&items, item)
 		}
 
+	case SymbolBitSetValue:
+		list.isIncomplete = false
+
+		enumv, ok := unwrap_bitset(ast_context, selector)
+		if !ok { break }
+
+		range, rok := get_range_from_selection_start_to_dot(position_context)
+		if !rok { break }
+
+		range.end.character -= 1
+
+		variable, vok := position_context.selector.derived_expr.(^ast.Ident)
+		if !vok { break }
+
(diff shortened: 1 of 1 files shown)


src/server/completion.odin
```diff
@@ -790,114 +844,112 @@ get_implicit_completion :: proc(
 
 	//infer bitset and enums based on the identifier comp_lit, i.e. a := My_Struct { my_ident = . } 
 	if position_context.comp_lit != nil {
-		if position_context.parent_comp_lit.type == nil {
```
laytan, line 793: This whole hunk/part of the diff is just turning this early return if into a regular if (in order to get to the part where call expressions are completed later). All this is just an indentation change.

DanielGavin:
Thanks for the PR.

PR #287, opened 2023-12-25
Major improvements to the textmate grammar

thetarnav:
I've been making a lot of small changes to the grammar since last time, but only now I think they are stable enough to share.

More and more tokens have their proper scopes guessed from just the syntax. I think that almost every token could be guessed accurately, with few exceptions to leave to semantic tokens.

Here are few changes:

- `typeid` is now a `support.type.primitive`, instead of a keyword, similar to other builtin types
- Added `$` and `%` as operators
- every type annotation (`foo: Bar(Baz)`) will assume only types can be on the right side
- assignments using `::` that aren't followed by `strict`, `proc`, etc. will have a `variable.other.constant` scope

image

- added block labels
- Every function call now has a `meta.function-call` scope.

image

- mark as types some tokens that can only be types

image

And few small fixes and improvements, like using `---` in forgein code blocks.

image

I've been vscode with those changes for a while and tried to go over odin and old codebases to see if there are some bugs with highlighting, but it seems to be working pretty well.
But it probably would be smart for someone else to try them as well.

### DIFF
--- a/editors/vscode/syntaxes/odin.tmLanguage.json
+++ b/editors/vscode/syntaxes/odin.tmLanguage.json
@@ -2,123 +2,182 @@
 	"comment": "Modification of https://github.com/fivemoreminix/odin-vscode",
 	"$schema": "https://raw.githubusercontent.com/martinring/tmlanguage/master/tmlanguage.json",
 	"name": "Odin",
-	"patterns": [{ "include": "#statements" }],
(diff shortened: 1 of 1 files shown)


DanielGavin:
Sweet thanks.

I'll try and get it on my computer locally and use it for a while.

PR #300, opened 2024-01-24
Remove where-clause from syntax

thetarnav:
Fixes a grammar bug that it seems I've introduced in the last change.
The where clause was the problem so now I removed it.

from
image

to
image

btw the extension is still using older grammars, not sure why
I needed to build locally to use the one in master


DanielGavin:
Strange it's not being built into the extension.

PR #333, opened 2024-04-01
removed duplicate condition

IsaiahPaget:
Hey there Daniel, I am F9pop from the Discord server. I found this duplicate condition, I removed it and all the tests still passed. Let me know what you think, cheers!

### DIFF
--- a/src/server/requests.odin
+++ b/src/server/requests.odin
@@ -207,11 +207,6 @@ read_and_parse_header :: proc(reader: ^Reader) -> (Header, bool) {
 			header.content_length = value
 
 			found_content_length = true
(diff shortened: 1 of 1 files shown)


DanielGavin:
```
if strings.compare(header_name, "Content-Length") == 0 {

} else if strings.compare(header_name, "Content-Type") == 0 {

}
```

One is length and the other is type. Doesn't seem like a duplicate to me. 

Also the tests do not test the actual request messages. Only the completion, hover, gotos, etc.

You will have to test that manually through your editor.

IsaiahPaget:
Whoops, I jumped the gun sorry about that.

DanielGavin:
No worries. Happens.

PR #351, opened 2024-04-15
support bit_fields

laytan:
They are pretty similar to structs (just a bit simpler) so I piggy backed of that a lot here, added some basic tests and tested the formatting myself.


DanielGavin:
Thanks for the work

PR #354, opened 2024-04-17
Fix check missing newlines

Feoramund:
The check procedure was missing newlines because `scan` was doing a double `advance`, causing multiple errors to appear in one program line. Using `next` instead should fix that.


DanielGavin:
Closing this because I have moved over to the json parsing, so the older code is not needed anymore.

PR #359, opened 2024-04-18
Emit proc "receiver" when using fake methods.

harold-b:
This allows procedure receivers to be emitted as the `TextEdit` when using fake methods. It will also add referencing or dereferencing as needed by the selector expression and target receiver type.


harold-b:
`force-push`: Rebased from latest `master`, had not merged it locally.

PR #376, opened 2024-05-03
Add a symbol for explicit `enum.member` selectors

thetarnav:
I noticed that `Enum.Enum_Member` expressions weren't highlighted properly.
Turns out that they weren't added to the symbols map, unlike `package.value` and others.
So I "fixed" that.
But returning a `SymbolEnumValue` for the member is probably wrong, it may need it's own symbol value. Especially since go to definition goes to the whole declaration, not specific member.
I can switch it to "patch" it in semantic_tokens, instead, similarly to how implicit selectors are handled. But thats also wrong, because `.Not_Actually_A_Enum_Member` shouldn't get the semantic highlighting.

I think that ideally, the whole semantic tokens functionality should just be handled by looping over the symbols map and writing those, instead of walking the ast, which was already walked by the analyzer.

Also removed the `selector` field from builder, it looks unused now.

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -1156,6 +1156,10 @@ internal_resolve_type_expression :: proc(
 				} else {
 					return Symbol{}, false
 				}
+			case SymbolEnumValue:
+				// enum members probably require own symbol value
+				selector.type = .EnumMember
+				return selector, true
 			}
 		}
 	case:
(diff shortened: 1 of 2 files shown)


DanielGavin:
Yeah potentially we need a symbol for the enum value. Yes the symbol resolving for references and semantic tokens should eventually be enough that we could just walk the symbols. And just leave everything else to the client syntax highlighter.

PR #378, opened 2024-05-04
Don't make a symbol for `nil`

thetarnav:
Alternative for #377 since a symbol for nil doesn't have much value now.

Don't make a symbol for `nil` and enable semantic tokens for `SymbolBasicValue`


DanielGavin:
Currently prefer nil not having a type.

PR #384, opened 2024-05-08
Add a .Type SymbolType enum

thetarnav:
Add a .Type SymbolType enum

Not sure if `Type` is the best name for something in a `SymbolType` enum,
But it's meant to be a generic variant for all types, besides the ones separately specified.
Basically for all the maps, arrays, slices, dyn arrays, matrixes, etc
To not have a symbol for each individual type, but a single one for all of them.
Basically everything that wasn’t covered by existing enum, but is a type, not just any constant.

Also I cleaned up the `visit_ident` proc, to have a clear split between handling types and variables.

before_after

now all the types are consistently correct
and all the constants consistently broken
but fixing that requires more work (and deeper changes)

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -2659,7 +2659,7 @@ make_symbol_array_from_ast :: proc(
 ) -> Symbol {
 	symbol := Symbol {
 		range = common.get_token_range(v.node, ast_context.file.src),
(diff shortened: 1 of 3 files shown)


DanielGavin:
I'll take a look at the constants when working on the new rename branch.

PR #385, opened 2024-05-09
Fix variable names in value declarations not having symbols

### LINKED ISSUE #374: Bug with nested declaration symbol shadowing
It seems that when shadowing a variable, the symbol for the returned value, is the symbol of the shadowed variable.

Kinda hard to explain

But this is what happens when doing "go to definition" for each `foo`:

```odin
foo :: 1         // no refs found; should go to itself

main :: proc() {
    foo := 2     // goes to 1    ; should go to itself

    foo          // goes to 2    ; good

    {
        foo := 3 // goes to 2    ; should go to itself

        foo      // goes to 3    ; good
    }
}
```

this also affects semantic tokens now as the tokens for identifiers on lhs are resolved using symbols.

also this case is weird:

```odin
main :: proc() {
    foo :: 1     // no refs found; should go to itself
    
    foo          // goes to 1    ; good
    
    {
        foo := 2 // goes to 1    ; should go to itself
    
        foo      // goes to 1    ; should go to 2
    }
}
```

thetarnav:
I was trying to fix a bug where local variables did not get any symbols in their own declarations.
e.g. in `foo := 123`, `foo` would get a symbol if it was declared globally, but not when declared in a proc scope.

The reason was that the offset of the `foo` ident in the decl was too low. The `get_local` proc only matched idents with locals if they were after the declaration.
So I added another hacky condition to match offsets with the `lhs` expression directly.
Kinda like matching pointers I guess.
But I couldn't use a pointer because `ast.Ident` was getting passed around as a value instead.

This solution doesn't feel correct because it is too indirect. I was thinking I could just use a simple `is_in_declaration` flag, to simplify it, but it turns out the walker in analysis doesn't even know if the ident is in a declaration or not?

This also partially fixes #374
But only the case where both the shadowed and shadowing variables are mutable.

Also `get_local` had some weird conditions in it
`i < 0` couldn't ever happened
and this one:

```odin
ret := local_stack[i].rhs
if ident, ok := ret.derived.(^ast.Ident);
	ok && ident.name == name {
	if i - 1 < 0 {
		return {}, false
	}
}
```

I'm guessing it has to do with doing `foo := foo`?
But I couldn't ever hit it.
Especially when it required `i` to be `0`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,6 @@
 *node_modules
 /ols
 /odinfmt
+tests/tests
 ols.json
 .vscode/settings.json
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -1219,31 +1219,22 @@ clear_local_group :: proc(ast_context: ^AstContext, id: int) {
 }
 
 get_local :: proc(
-	ast_context: ^AstContext,
-	offset: int,
-	name: string,
+	ast_context: AstContext,
+	ident: ast.Ident,
 ) -> (
 	DocumentLocal,
 	bool,
 ) {
-	for _, locals in &ast_context.locals {
-		if local_stack, ok := locals[name]; ok {
-			for i := len(local_stack) - 1; i >= 0; i -= 1 {
-				if local_stack[i].offset <= offset ||
-				   local_stack[i].local_global {
-					if i < 0 {
-						return {}, false
-					} else {
-						ret := local_stack[i].rhs
-						if ident, ok := ret.derived.(^ast.Ident);
-						   ok && ident.name == name {
-							if i - 1 < 0 {
-								return {}, false
-							}
-						}
-						return local_stack[i], true
-					}
-				}
+	for _, locals in ast_context.locals {
+		local_stack := locals[ident.name] or_continue
+
+		#reverse for local in local_stack {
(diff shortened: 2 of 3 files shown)


DanielGavin:
You could potentially also look at: get_locals_value_decl.

Change what the offset is of the local when it's being stored.

The whole walker was designed for completion, it is being used for something it wasn't really planned for - semantic tokens. Now it's being asked what the symbol is of something being declared, which is meaningless question for completion.

DanielGavin:
image

Semes to work in vscode without any code changes.

thetarnav:
Simply changing the offset won't work here, right? The third `foo` is referencing the first local, not second.
```odin
foo := 123
{
    foo := foo
}
```

DanielGavin:
Most likely not. I thought you had issues with `foo := 2`, but that works for me.

DanielGavin:
Might just allow this hack. It might also be possible to check the identifier is in `ast_context.value_decl`, it's only initialized in the completion. But could also be set in hover and goto. Then just check if the identifier is in. This case can only happen in the beginning, since the first check will be the declaration or never again.

DanielGavin:
Merging it for now. We have some tests now for shadowing, so could be helpful if deciding to change it.

PR #395, opened 2024-06-01
Add `unreachable` to `builtin/builtin.odin`

thetarnav:
was recently added to odin in https://github.com/odin-lang/Odin/commit/451dc645df9f12058267d7d5ae0f8ac8a1fb8998

btw, why do some procs in `builtin.odin` have `@builtin` attr, and some don't?


DanielGavin:
I assume it's just something that has been forgotten to be added. I'm not even sure if the ones that haven't got it, should be in builtin.

PR #405, opened 2024-06-13
Semantic tokens tests

thetarnav:
Added setup for testing semantic tokens and add one simple test for starters.

But that one test has already revealed some bug that a procedure identifier is receiving both function and type tokens. I wasn't able to find a quick fix yet that didn't break proc type decls.
But the type token is ignored so it's all displayed correctly.

The pr is based on the rename branch to not go through merge conflicts

I'll be adding more test cases later


DanielGavin:
Great work with adding semantic token tests. Was thinking of adding that eventually.

PR #407, opened 2024-06-18
add additional completion snippets for dynamic arrays

hhhapz:
This commit adds a handful of completion snippets for dynamic arrays from the builtin package.

Some of these are less often used than others, however since the completion entries otherwise should always be empty, I think that filling in all of these should be OK.

The most useful one for me, personally is `append`, however I felt like while I was at it, I might as well add the rest as well :).

### DIFF
--- a/src/server/completion.odin
+++ b/src/server/completion.odin
@@ -1947,6 +1947,56 @@ append_magic_dynamic_array_completion :: proc(
 		append(items, item)
 	}
 
+	// This proc is shared between slices and dynamic arrays.
+	if _, ok := symbol.value.(SymbolDynamicArrayValue); !ok {
+		return
+	}
+
+	//pop
+	{
+		text := fmt.tprintf("pop(&%v)", symbol_str)
+
+		item := CompletionItem {
+			label = "pop",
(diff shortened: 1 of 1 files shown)


DanielGavin:
Yeah it's fine to add them as snippets. 

There seems to be one issue:

```
newText = fmt.tprintf("%s(&%v, $0)", name, symbol_str),
```
You assume that the array is not a pointer by hardcoding `&`.

It might have to be something like this:

```
if symbol.pointers > 0 {
  common.repeat(
  "^",
  symbol.pointers-1, //If it's zero it just doesn't do anything
  context.temp_allocator,
  )
} else {
  just hardcode &
}

```

Possible arrays: `^[dynamic]int, ^^[dynamic]int, [dynamic]int`

hhhapz:
good catch

hhhapz:
@DanielGavin I'm still relatively new odin, and the lack of `+` was painful. Let me know if you feel there's a more 'idiomatic' way to do this that I might have missed. This seemed the most clear and straightforward to me.

DanielGavin:
Seems good now.

PR #418, opened 2024-06-30
Add `raw_data` proc to `builtin/builtin.odin`

### LINKED ISSUE #417: Add support for `raw_data` macro
The `raw_data` macro is too complicated to be added to `builtin.odin` probably.
So it needs to be hard coded, but it would be nice to have a autocomplete and hover information about it and the returned types.
in odin code its something like this:
```odin
raw_data_string  :: proc (v: string)         -> [^]u8                           ---
raw_data_slice   :: proc (v: $T/[]$E)        -> [^]E                            ---
raw_data_dynamic :: proc (v: $T/[dynamic]$E) -> [^]E                            ---
raw_data_array   :: proc (v: ^$T)            -> [^]intrinsics.type_elem_type(T) ---
```

thetarnav:
Fixes #417

This works, but not without issues.
it does autocomplete, but it also

### DIFF
--- a/builtin/builtin.odin
+++ b/builtin/builtin.odin
@@ -34,6 +34,14 @@ conj       :: proc(value: Complex_Or_Quaternion) -> Complex_Or_Quaternion ---
 
 @builtin unreachable :: proc() -> ! ---
 
+@private _raw_data_slice   :: proc(value: []$E)         -> [^]E    ---
+@private _raw_data_dynamic :: proc(value: [dynamic]$E)  -> [^]E    ---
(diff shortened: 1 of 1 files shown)


DanielGavin:
I think it works in rename branch.

thetarnav:
And do you know why the private procs are showing up? it is because builtin is being added to the „global scope” or something?

DanielGavin:
Have you tried doing `private="file"`?

thetarnav:
it does indeed help with autocomplete, but for some reason it completely disables hover info

DanielGavin:
I'll have a look at that later.

PR #426, opened 2024-07-11
fix resolve poly for proc types

laytan:
You can consider this a bug report with an attempted fix from myself.

Opening this would crash (I think you need to have inlay hints enabled):
```odin
package main

main :: proc() {
    b := foo(1, 2, proc(a: int, b: int) {})
}

foo :: proc {
    foo1,
    foo2,
}

foo1 :: proc(a: $T, c: $C/proc(a: T)) -> int {
    return 1
}

foo2 :: proc(a: $T, b: $T2, c: $C/proc(a: T, b: T2)) -> int {
}
```

from reading the other cases this seems the appropriate fix but I could be wrong.

Inlay hints with procedure groups don't work anyway so this doesn't actually change anything noticeable, just prevents a crash.

### DIFF
--- a/src/server/generics.odin
+++ b/src/server/generics.odin
@@ -143,7 +143,6 @@ resolve_poly :: proc(
 				}
 			}
 		}
-	case ^ast.Struct_Type:
 	case ^ast.Dynamic_Array_Type:
 		if call_array, ok := call_node.derived.(^ast.Dynamic_Array_Type); ok {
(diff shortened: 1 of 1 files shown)


DanielGavin:
I'll check it out. I didn't even know you could use generics with procedure signatures like that. That complexes everything further with group in generics to find the correct candidates.

laytan:
Yeah I originally got the pattern from `core:thread`, it has `_with_data` procs using this, only when I added it to a proc group OLS started crashing.

PR #452, opened 2024-07-28
Correct collecting symbols for array types

### LINKED ISSUE #449: Array type aliases work differently between files
```odin
// a.odin
Vec :: [2]f32 // Vec shows up as [2]f32
foo :: proc (v: Vec) {} // here too, v as well


// b.odin
main :: proc () {
	v: Vec // Vec and v shows up as []f32
}
```

image

there is also a difference in syntax highlighting, which suggests that ols thinks that `Vec` in the second file is a value of type `[]f32`, and not a type alias.
It only happens with array type aliases.

Might be related to #345

thetarnav:
Fixes #449

Turns out that the `collect_array` and `collect_slice` had swapped bodies.

It wasn't noticeable in the same file because variables from the same file were found in `ast_context.globals` which doesn't use symbols, where variables from other files are resolved by `memory_index_lookup` which uses symbols.

I also changed the symbol types for other type definitions to `.Type` from `.Variable`.

Also corrected the way hover strings are being tested, from `strings.contains` to exact match. I wanted to add a test case for the fix, but that would require to change more about the test setup.

### DIFF
--- a/src/server/collector.odin
+++ b/src/server/collector.odin
@@ -313,6 +313,22 @@ collect_slice :: proc(
 	collection: ^SymbolCollection,
 	array: ast.Array_Type,
 	package_map: map[string]string,
+) -> SymbolSliceValue {
+	elem := clone_type(
+		array.elem,
+		collection.allocator,
+		&collection.unique_strings,
+	)
+
+	replace_package_alias(elem, package_map, collection)
+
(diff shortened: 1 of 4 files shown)


DanielGavin:
Yeah testing currently only supports testing one file, but with multiple packages. So it's not really possible to test multiple files in one package.

PR #455, opened 2024-07-29
Add missing op token when building string for unary expr node

### LINKED ISSUE #345: `[?]` arrays are displayed as slices on hover
```odin
numbers :: [?]int{1, 2, 3}
```
`numbers` is displayed as `[]int`, but should be `[?]int` or `[3]int`

thetarnav:
Fixes #345 

It would be cool to actually infer the length of the array in the future and display it as `[4]` instead of `[?]` but that doesn't look too easy by how the printing is currently setup.

### DIFF
--- a/src/common/ast.odin
+++ b/src/common/ast.odin
@@ -1086,6 +1086,7 @@ build_string_node :: proc(
 	case ^Tag_Expr:
 		build_string(n.expr, builder, remove_pointers)
 	case ^Unary_Expr:
+		strings.write_string(builder, n.op.text)
 		build_string(n.expr, builder, remove_pointers)
 	case ^Binary_Expr:
 		build_string(n.left, builder, remove_pointers)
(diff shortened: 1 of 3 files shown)


thetarnav:
writing the op token for the unary operator has caused a crash because of a seg fault for me now. trying to debug. it happened during loading completions, when making a string for the `shrink_dynamic_array` proc. One of its params have a `-1` as a default value, and printing that was causing the seg fault.
It can be simply reproduces by typing `s` anywhere in this branch.
Looks like the file string for the token was maybe dealocated? it's showing in the debugger as `{data='\0', len=51}` which doesn't make sense.

DanielGavin:
It's most likely because the text has not been interned and is garbage.

Example:
```
	case ^Binary_Expr:
		n := node.derived.(^Binary_Expr)
		r.left = clone_type(r.left, allocator, unique_strings)
		r.right = clone_type(r.right, allocator, unique_strings)
		//Todo: Replace this with some constant table for operator text
		if unique_strings == nil {
			r.op.text = strings.clone(n.op.text, allocator)
		} else {
			r.op.text = get_index_unique_string(unique_strings, allocator, n.op.text)
		}
```

But there is none in unary expr:

```
	case ^Unary_Expr:
		r.expr = clone_type(r.expr, allocator, unique_strings)
```

thetarnav:
yup, that was it

PR #466, opened 2024-08-05
Add documentation for enabling references and fix enable_rename config flag

BradLewis:
Whilst I was playing around with odin and ols in neovim, I noticed I was unable to go-to-references as neovim was saying the "textDocument/references" was unsupported for the server, even though the documentation seemed to indicate this should be supported.
After a bit of investigation, I realised that "textDocument/references" is actually behind a feature flag.

So this simple PR is just to document that in the README.

I also noticed that the `enable_rename` flag was actually using `ols_config.enable_references`. Not sure if this was intentional, but I also fixed that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,10 @@ Options:
 
 `enable_checker_only_saved`: Turns on only calling the checker on the package being saved. 
 
+`enable_references`: Turns on finding references for a symbol.
+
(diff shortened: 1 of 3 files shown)


DanielGavin:
It was not intended with the enable_rename flag.

PR #471, opened 2024-08-12
short assignments dont get a new line in `case:`

asibahi:


DanielGavin:
Not sure If I want this to be default. It's a fairly simple thing, so adding a config for it in the branch statement is good enough.

There are issues with consistency when you have one case with simple statements then with statements more than one statement.


```
switch v {
case .One
    my_value -= 1
    my_value_2 += 2   
case .Two: my_value += 1
case . Three: 
    my_value -= 2
    my_value_2 += 2
}
```

thetarnav:
Maybe it should check if all cases are one-liners
so its either
```odin
switch v {
case .One my_value -= 1
case .Two: my_value += 1
case .Three: my_value -= 2
}
```
or
```odin
switch v {
case .One
    my_value -= 1
    my_value_2 += 2
case .Two:
	my_value += 1
case .Three: 
    my_value -= 2
    my_value_2 += 2
}
```

DanielGavin:
It's more consistent to do either everything or nothing.

PR #492, opened 2024-08-30
Use a new file tag parser

thetarnav:
A proper file tag parser was just added to odin so this uses it instead of current parsing method.
Also simplified `collect_value_decl` as bit


DanielGavin:
Thanks.

PR #493, opened 2024-08-30
Fix import issues

### LINKED ISSUE #443: @(require) on import is Confused on How to Format Correctly
```f#
package main

@(require)import "core:os"

main :: proc() {}
```
save
```f#
package main

@(require)
import "core:os"

main :: proc() {}
```
save
```f#
package main

@(require)import "core:os"

main :: proc() {}
```
...etc

### LINKED ISSUE #486: odinfmt deletes imports that are at the very bottom of the file
to repro, add some imports at the bottom of an odin file and format it

e.g. I can repro with https://github.com/odin-lang/Odin/blob/master/vendor/glfw/wrapper.odin, I just copied the imports at the top, pasted them below everything else and ran `odinfmt wrapper.odin`

laytan:
Fixes #443 
Fixes #486

### DIFF
--- a/src/odin/printer/printer.odin
+++ b/src/odin/printer/printer.odin
@@ -273,6 +273,12 @@ print_sorted_imports :: proc(p: ^Printer, decls: []^ast.Stmt) {
 		decl.pos.line = start_line + i
 		decl.end.line = start_line + i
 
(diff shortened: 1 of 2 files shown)


DanielGavin:
Thanks!

PR #500, opened 2024-09-05
Fix Emacs configuration for Doom Emacs compatibility

jtsiros:
Ensure `lsp-language-id-configuration` is properly initialized by updating the configuration to load after `lsp-mode`. This prevents potential `(void-variable lsp-language-id-configuration)` errors caused by referencing the variable before `lsp-mode` is loaded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,14 +214,22 @@ lspconfig.ols.setup({})
 ### Emacs
 
 ```elisp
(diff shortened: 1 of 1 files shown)


DanielGavin:
Thanks

elken:
While this helps, it's not clear in the instructions that they're specific to Doom Emacs. Especially since it's using `with-eval-after-load` instead of `after!` I think the `use-package!` should be `use-package` 😄

DanielGavin:
@elken  Go ahead. Would be useful.

Issue #489, opened 2024-08-24
OLS not usable offline / nightly updated fails / seemingly breaking the extension


michaelhartung:
Hi,

I launched VSCode while not connected to the internet and started editing a `.odin` file.

OLS Console showed that it failed to fetch latest nightly build:

````
INFO [24/08/2024, 10:34:53]: PersistentState: {
  lastCheck: 1724277627007,
  releaseId: 170369909,
  serverVersion: '0.1.30'
}
DEBUG [24/08/2024, 10:34:53]: Issuing request for released artifacts metadata to https://api.github.com/repos/DanielGavin/ols/releases/tags/nightly
ERROR [24/08/2024, 10:35:04]: Bootstrap error [FetchError: request to https://api.github.com/repos/DanielGavin/ols/releases/tags/nightly failed, reason: getaddrinfo ENOTFOUND api.github.com
	at ClientRequest.<anonymous> (/Users/michael/.vscode/extensions/danielgavin.ols-0.1.30/node_modules/node-fetch/lib/index.js:1491:11)
	at ClientRequest.emit (node:events:531:35)
	at TLSSocket.socketErrorListener (node:_http_client:500:9)
	at TLSSocket.emit (node:events:519:28)
	at emitErrorNT (node:internal/streams/destroy:169:8)
	at emitErrorCloseNT (node:internal/streams/destroy:128:3)
	at process.processTicksAndRejections (node:internal/process/task_queues:82:21)] {
  type: 'system',
  errno: 'ENOTFOUND',
  code: 'ENOTFOUND'
}
````

Attempts to restart OLS result in `ols.start` command not found errors. Had to re-connect to the internet and re-install the extension to get it working again.

<img width="352" alt="image" src="https://github.com/user-attachments/assets/5b7c147f-5514-4a2e-89f3-abe1dfdf3697">

michaelhartung:
sorry for the noise, realised this is the wrong repo, apologies

michaelhartung:
sorry for the noise, realised this is the wrong repo, apologies

PR #534, opened 2024-10-22
Allow to dismiss update dialog and continue to work offline

dayllenger:
If nightly check failed because of simply being offline, the language server won't start. This patch allows to dismiss the popup. If nightly check passed but download failed, then it's not going to start.

related #489

### DIFF
--- a/editors/vscode/src/extension.ts
+++ b/editors/vscode/src/extension.ts
@@ -71,7 +71,7 @@ export async function activate(context: vscode.ExtensionContext) {
     );
 
     context.subscriptions.push(disposable);
- 
+
     if (!serverPath) {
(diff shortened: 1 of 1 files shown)


dayllenger:
By the way, typescript dumps a bunch of errors like this:
```
node_modules/vscode-languageclient/lib/common/client.d.ts:1:838 - error TS2305: Module '"vscode"' has no exported member 'InlineCompletionItemProvider'.

node_modules/vscode-languageclient/lib/common/codeConverter.d.ts:73:98 - error TS2694: Namespace '"vscode"' has no exported member 'InlineCompletionContext'.
```
It's old `@types/vscode` or something.

PR #557, opened 2024-12-29
fix proc returning proc with where constraint

sheganinans:
Currently ols will replace a line such as:

```odin
foo :: proc(x: $T) -> (proc(_: bool) -> bool) where intrinsics.check(T) {
  ...
}
```

with:

```odin
foo :: proc(x: $T) -> proc(_: bool) -> bool where intrinsics.check(T) {
  ...
}
```

Causing a parse error.

Here is an initial commit to bikeshed a solution. I haven't been able to quite test this fix yet, but (hopefully) it will work as expected.

I'm sure that a test case would be expected to be added to the test suite, but I'm not sure the best place to add it.

Just putting this out for comments/criticisms.

Thanks!

Aistis Raulinaitis

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -1648,7 +1648,7 @@ visit_expr :: proc(
 			document = cons(document, text("#force_no_inline"))
 		}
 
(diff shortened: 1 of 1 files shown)


DanielGavin:
Thanks! Yeah it seems to work fine.

PR #560, opened 2025-01-01
Add ols version

shizeeg:
```sh
% ./ols version
ols version dev-2024-11-9:g584f01b
```
Now  prints out it's version to be consistent with `odin version`
what do you think?


DanielGavin:
Yeah seems like a useful thing to have.

PR #571, opened 2025-01-07
Update Emacs documentation

elken:
Include some extra information, add steps for eglot and clean up the
suggested code.

Marked as a draft for now as I've submitted a [patch](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=75421) to have OLS included upstream. Depending on whether it goes in or not, I'll have to update what's here; but I've raised this now anyway so you can review :smile:


elken:
The aforementioned patch was merged in so as of Emacs 30 setup becomes much simpler 😄

DanielGavin:
Thanks!

PR #577, opened 2025-01-10
Fix make `ols version` work with GNU sed and BSD/macOS sed

shizeeg:
since GNU sed and BSD sed behave differently, update VERSION using a temporary file.

tested on Arch Linux and FreeBSD 14.1

P.S. oops, my bad.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -48,6 +48,6 @@ fi
 
 version="$(git describe --tags --abbrev=7)"
 version="${version%-*}:${version##*-}"
-sed -i "" "s|VERSION :: .*|VERSION :: \"${version}\"|g" src/main.odin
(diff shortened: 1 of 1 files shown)


DanielGavin:
Don't think that will run on MacOS.

https://github.com/DanielGavin/ols/pull/572

You probably need to add '-i' flag to it.

Maybe  @glektarssza can confirm whether the new changes will work on MacOS

glektarssza:
This does run on macOS. Tested on my local system. It's simply a different way of fixing the issue I fixed in #572. If this works on Linux too then I think this is probably the better fix.

DanielGavin:
Great merging then. Thanks.

PR #584, opened 2025-01-19
add special coloring for all caps variables

Goldenlion5648:
I use the convention of all caps for variables that I declare as constant, so I like having special coloring for those. In my color scheme ("monokai" in vscode):

Before:
image

After:
image


Goldenlion5648:
any update one this?

PR #585, opened 2025-01-19
Fix extra new lines before brace when using Allman brace style

### LINKED ISSUE #530: odinfmt: "Allman" brace style inserts extra new lines in switch
So far this issue specifically happens with `switch` statements, after every format command issued, the odinfmt adds an extra line before the curly bracket opens until it reaches maximum of 3 new lines and then stops.

Expected behavior of "Allman" brace style:
```odin
num := 1

switch num
{
        case 0:
        case 1:
}
```

Current behavior:
```odin
num := 1

switch num


 
{
        case 0:
        case 1:
}
```

henbr:
This fixes #530 

It also changes so the `odinfmt.json` is used for the `odinfmt` tests. The  `odinfmt.json` is read from the same dir as the file under test, if there is no config file default config will be used. This makes it possible to have different configs for each test.

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -1042,7 +1042,9 @@ visit_stmt :: proc(
 		}
 
 		document = cons_with_opl(document, visit_expr(p, v.cond))
+		set_source_position(p, v.body.pos)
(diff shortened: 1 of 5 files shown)


DanielGavin:
Thanks. Yeah good idea to also add test directory for allman.

PR #588, opened 2025-02-02
More magic completion

### LINKED ISSUE #573: Missing `len` fake-method for arrays and strings
Related: #61

Dynamic arrays and slices have a `thing.le -> len(thing)` autocomplete, but not static arrays and strings.

```odin
arr:     [3]int       // X
dyn_arr: [dynamic]int // ✓
slice:   []int        // ✓
str:     string       // X
```

cornishon:
The missing builtin procedures taken from "runtime/core_builtin.odin".

Also fixes #573


DanielGavin:
Thanks!

PR #611, opened 2025-03-04
Put spaces around braces of single line blocks

sndb:
Add an option to put spaces around braces of single line blocks.
For example, `if true {return 0}` => `if true { return 0 }`.


BigBoyBarney:
Is there a blocker for this? Would be a really nice QoL change :D

BradLewis:
It looks good to me, I'll merge it in.

BigBoyBarney:
Thank you! :heart:

PR #613, opened 2025-03-09
Use odin_command from ols.json to check if odin is installed & allow relative path for odin_command

hardliner66:
Changes:
- ols.exe
  - use the correct odin binary when specified as a relative path
- vscode extension
  - use odin_command to determine if the odin binary exists

I know, this should in reality be two separate pull requests, but I realized too late that I had to touch both the extension and the ols source for this to work properly (and without errors).

If you want I can still split it into two pull requests.

### DIFF
--- a/editors/vscode/src/config.ts
+++ b/editors/vscode/src/config.ts
@@ -3,51 +3,52 @@ import { log } from "./util";
 
 export class Config {
 
-    readonly extensionId = "danielgavin.ols";
-    readonly rootSection = "ols";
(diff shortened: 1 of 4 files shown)


762SPR:
@DanielGavin FYI this does not seem to have fixed the prompt in VSCode that complains Odin cannot be found as described in #624 

Just a minor annoyance at this point as the server does seem to be working, but it still complains that it can't find it when you first open the file. Version 0.1.34, VSCodium, Arch Linux (Endeavor)

Issue #616, opened 2025-03-23
odinfmt segfault on Linux


drshapeless:
```odin
./odinfmt -stdin
```
It will return this.
```
zsh: segmentation fault  ./odinfmt -stdin
```

If I do this.
```odin
./odinfmt -w src/main.odin 
```
It cannot read anything
```
 is neither a directory nor a file 
```

I tried to build a debug build, here is the backtrace from lldb.
```
lldb odinfmt 
(lldb) target create "odinfmt"
Current executable set to '/home/jacky/src/ols/odinfmt' (x86_64).
(lldb) r -stdin 
Process 2256 launched: '/home/jacky/src/ols/odinfmt' (x86_64)
Process 2256 stopped
* thread #1, name = 'odinfmt', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x0)
    frame #0: 0x00007ffff7e7d744 libc.so.6`___lldb_unnamed_symbol4479 + 68
libc.so.6`___lldb_unnamed_symbol4479:
->  0x7ffff7e7d744 <+68>: movb   (%rsi), %cl
    0x7ffff7e7d746 <+70>: je     0x7ffff7e7d752 ; <+82>
    0x7ffff7e7d748 <+72>: movzwl -0x2(%rsi,%rdx), %esi
    0x7ffff7e7d74d <+77>: movw   %si, -0x2(%rdi,%rdx)
(lldb) bt
* thread #1, name = 'odinfmt', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x0)
  * frame #0: 0x00007ffff7e7d744 libc.so.6`___lldb_unnamed_symbol4479 + 68
    frame #1: 0x0000000000406065 odinfmt`odin_format::find_config_file_or_default [inlined] runtime::copy_from_string:proc"contextless"(dst:[]u8,src:string)->(:int) at core_builtin.odin:85:3
    frame #2: 0x0000000000406054 odinfmt`odin_format::find_config_file_or_default [inlined] strings::clone_to_cstring at strings.odin:64:2
    frame #3: 0x0000000000406054 odinfmt`odin_format::find_config_file_or_default [inlined] filepath::abs at path_unix.odin:26:2
    frame #4: 0x0000000000405fc9 odinfmt`odin_format::find_config_file_or_default at format.odin:22:14
    frame #5: 0x0000000000410e20 odinfmt`main [inlined] odinfmt::main at main.odin:107:2
[...]

PR #619, opened 2025-04-05
Check for path length

Akselmo:
On Linux, when running `odinfmt -stdin` it crashes `odinfmt`.

From what investigation I've done in this issue https://github.com/DanielGavin/ols/issues/616
it seems that the `path` variable is just `''` empty character. However, it counts as one character when checking with `len(path)`. Then whatever tries to process that string further in the chain gets upset about it.

I am not entirely sure if this is bug in Odin upstream or in ols, but having this check seems to fix it. I think having the check in general outside of the `strings.has_prefix` check here makes sense.

(First time contributing to Odin project, if there's something silly I've missed let me know :D )

### DIFF
--- a/tools/odinfmt/main.odin
+++ b/tools/odinfmt/main.odin
@@ -90,6 +90,9 @@ main :: proc() {
 	path := os.args[len(os.args) - 1]
 	if strings.has_prefix(os.args[len(os.args) - 1], "-") {
 		args_to_parse = os.args[1:]
+	}
+
+	if len(path) <= 1 {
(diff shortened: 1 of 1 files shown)


tools/odinfmt/main.odin
```diff
@@ -186,4 +189,4 @@ main :: proc() {
 	}
 
 	os.exit(1 if write_failure else 0)
-}
+}
```
Akselmo, line 192: No idea what this is about, I guess line endings?

DanielGavin:
I think it's related to Odin. But having this fix for now is fine.

Issue #644, opened 2025-06-01
Incorrect enum when goto_definition inside enum switch


Koliras:
`goto_definition` sends me to enum of switch case instead of the one I am trying to asign to variable if I am asigning inside the switch case.

Minimal reproduction:
```odin
package odin_test

Bar :: enum {
	Bar1,
	Foo1,
}

Foo :: enum {
	Foo1,
	Foo2,
}


main :: proc() {
	my_foo: Foo
	my_bar: Bar
	switch my_foo {
	case .Foo1:
		my_bar = .Foo1
	case .Foo2:
		my_bar = .Bar1
	}
}
```
When I try to go to definition of `.Bar1` on `my_bar = .Bar1` line, ols sends me to `Foo :: enum {` line.
If I try to go to `.Foo1` on line `my_bar = .Foo1`, ols sends me to `Foo1` inside the `Foo` enum, not the one in `Bar` enum

PR #647, opened 2025-06-05
Resolve assignments before switch statements position context

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/644.

When inferring enum assignment within a switch statement, the lsp would set the switch context as `switch_stmt`, see it was dealing with an enum and so assume that it's the enum being switched on rather than assigned to. Simple solution is to just handle assignments before `switch_stmt`.

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -1669,17 +1669,16 @@ resolve_implicit_selector :: proc(
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


DanielGavin:
Thanks for the recent commits!

BradLewis:
No worries! It's been fun

PR #648, opened 2025-06-06
Add hover support for implicit selector expr

BradLewis:
Add hover support to the implicit selector enum and union expressions as it was something I was doing to remind myself of the underlying enum type. 

I'm not sure what the desired value for this should be but I've gone with simply `<pkg>.<name>: .<value>` as that's generally what I'm after. Let me know what you think.

Currently for unions it only works if it's the union of enum types. I noticed this also applies to `textDocument/definition` so I might take a look at that at a later point. Edit: I had a quick look at this and it seems a very straightforward fix so I just added it to this PR.

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -3727,9 +3727,10 @@ unwrap_super_enum :: proc(
 
 	for type in symbol_union.types {
 		symbol := resolve_type_expression(ast_context, type) or_return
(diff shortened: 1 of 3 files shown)


DanielGavin:
Yeah would have also chosen: `<pkg>.<name>: .<value>`.

Seems good.

PR #650, opened 2025-06-08
Enrich hover information

BradLewis:
This PR is to enrich the information displayed by `textDocument/hover`, in particular with structs, enums and unions. It essentially adds the full definition similarly to what is seen in other languages. For example, for a simple program:
```odin
package main

T :: struct {
    a: int,
    b: string,
}

main :: proc() {
   t{*} := T{}
}
```
will show up in my IDE like
image
and similarly for unions and enums. 

I haven't spent much time cleaning it up yet, but it's been working well for me so it's good for some feedback if this is something you would like added to the LSP.

I could add this behind a configuration flag as well.

Thanks!

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -3765,7 +3765,7 @@ unwrap_bitset :: proc(ast_context: ^AstContext, bitset_symbol: Symbol) -> (Symbo
 	return {}, false
 }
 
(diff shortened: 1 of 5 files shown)


DanielGavin:
Seems to provide more information. It could later be added behind a configuration if people don't like it.

PR #655, opened 2025-06-10
Feat/enrich hover overload definition

BradLewis:
Adds the overloaded procedure definitions to the `textDocument/hover` information

As an example, looking at `strings.builder_make()`
Before:
image

After:
image


DanielGavin:
Yeah makes sense.

Also someone mentioned on discord they also wanted this feature for completion.

So before selecting the overloaded version they could quickly see the different procedures.

BradLewis:
Yeah I was thinking of going through and updating this kind of thing for completions as well. Right now I'm focusing on hover information as it covers the majority of needs for me.

PR #660, opened 2025-06-11
Fix issue with the LSP incorrectly inferring positions if the file started with an empty line

BradLewis:
Someone in the discord raised an issue where they were having weird goto definition and hover behaviour, where it would  take them to the symbol above them. They later realised it was due to them having an empty newline at the top of the file. After a quick look into it, it turns out when calculating positions it was starting from the 2nd character, rather than the 1st so it was skipping the newline.

### DIFF
--- a/src/common/position.odin
+++ b/src/common/position.odin
@@ -43,7 +43,7 @@ get_absolute_position :: proc(position: Position, document_text: []u8) -> (Absol
 	}
 
 	line_count := 0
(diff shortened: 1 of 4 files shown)


DanielGavin:
Nice. Surprised I haven't encountered that myself.

BradLewis:
Yeah I guess it's something that just rarely comes up.

Issue #659, opened 2025-06-11
Struct field completion not working for multi-pointer struct fields


nadako:
Something that was bugging me for a while with Vulkan API: if a struct has a field that is a multi-pointer to another struct, then the struct field completion doesn't work if you inline that field value, despite even having an explicit type in the literal (as Odin compiler requires).

```odin
package main

S1 :: struct {
	s2_ptr: [^]S2,
}

S2 :: struct {
	field: int,
}

main :: proc() {
	x := S1 {
		s2_ptr = &S2 { /* no field completion here */ }
	}
}
```

PR #665, opened 2025-06-13
Fix issue with field completions in multi-pointer structs and rename `SymbolMultiPointer`

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/659


nadako:
Damn that was fast, thanks so much!

PR #685, opened 2025-06-28
Improvements to local variable references

BradLewis:
At first I noticed when working in the code for `get_signature` (https://github.com/DanielGavin/ols/blob/master/src/server/documentation.odin#L21), I was unable to rename or find references to `builder` in the switch statements. I was able to easily fix that by just adding a `local_scope` for case_clauses, however after that if I added a `builder` variable before the switch statement, it would break again when referencing `builder` inside the switch statement, and when referencing the `builder` outside the switch statement if would match to all the `builder`s inside all of the cases. This also happened for all bodies as far as I could tell.

To fix this I've turned the `AstContext.locals` into a stack that we iterate in reverse order, rather than a map that gets iterated in a random order. After making that change, the `local_id` was no longer needed so I removed that. 

I'm not sure if there are some unintended consequences from this, but everything seems to be working as far as I can tell. And all of the nested reference cases are now working.

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -74,7 +74,6 @@ DocumentLocal :: struct {
 	offset:          int,
 	resolved_global: bool, //Some locals have already been resolved and are now in global space
 	local_global:    bool, //Some locals act like globals, i.e. functions defined inside functions.
-	id:              int, //Id that can used to connect the local to something, i.e. for stmt begin offset
 	pkg:             string,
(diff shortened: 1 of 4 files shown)


DanielGavin:
This change makes more sense than the current implementation.

BradLewis:
Perfect, I just wasn't sure if there wasn't anything that relied on the fact it was a map, but as far as I could tell that wasn't the case.

PR #689, opened 2025-06-29
feat: updated readme for ols instructions for lsp-mode

samwdp:
Updated the readme as I made a PR for ols to be included as a pre-configured source in emacs-lsp mode as of [4818](https://github.com/emacs-lsp/lsp-mode/pull/4818). This should be released in the next version of lsp-mode but I have included a way of installing from the master branch of lsp-mode using the straight.el package manager


DanielGavin:
Thanks!

Issue #693, opened 2025-06-29
No completion for small_array_slice


bangbangsheshotmedown:
```odin
import "core:container/small_array"

Data::struct {
    aaaaa:int,
}


MAX :: 4
a: small_array.Small_Array(MAX, Data)


testme::proc()
{
    for item in small_array.slice(&a) {

        ite // here no completion for item
    }
}
```

I compiled latest version of ols, same issue

bangbangsheshotmedown:
Thanks for the quick fix

PR #696, opened 2025-06-30
Improve resolution of parameteric types with external types

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/693

I'm just not sure what the `struct_value.poly == nil` check was there for originally, so I'm not sure if there's something else that would break with this.


DanielGavin:
I don't know why it's there. But if the tests don't fail, then it's probably good enough. Would hope that I added tests if it was important.

PR #712, opened 2025-07-03
Add //ols:private directive to skip parsing private sections of docum…

xoxorwr:
Odin’s import system is convenient since you don’t have to specify imports within a folder, but it results in LSP completion being cluttered with too many symbols, which make browsing difficult.

Using @(private) on each symbol isn’t ideal either, because these sections aren’t truly private API, I just want to exclude certain sections of files from LSP parsing.

### DIFF
--- a/src/server/documents.odin
+++ b/src/server/documents.odin
@@ -410,7 +410,8 @@ parse_document :: proc(document: ^Document, config: ^common.Config) -> ([]Parser
 		pkg.kind = .Runtime
 	}
 
-	doc_end := document.used_text if document.private_scope == -1 else document.private_scope
+	doc_end = document.used_text if document.private_scope == -1 else min(document.private_scope, document.used_text)
+
 	document.ast = ast.File {
 		fullpath = document.fullpath,
 		src      = string(document.text[:doc_end]),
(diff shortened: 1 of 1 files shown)


BradLewis:
I think this is something that should be handled on the client side. The lsp is there to provide you with the symbols within the workspace/file. With that information you can sort and filter these entries as you wish.

xoxorwr:
I lack knowledge unfortunately, I don't know how to achieve this, do you have an idea?

I really need this because I have plenty of bindings, and they literally bloat the completion popup

DanielGavin:
It could be a bug. Are you saying that procedures are shown when called from different packages? As in you say `my_binding_package.`, and it then shows procedures in files that should be private?

xoxorwr:
This isn’t a bug, it's to make the experience nicer and less convoluted when I'm working on code that's in a module with bindings, there is no namespace in Odin unfortunately, and no procs inside structs either

BradLewis:
> I lack knowledge unfortunately, I don't know how to achieve this, do you have an idea?
> 
> I really need this because I have plenty of bindings, and they literally bloat the completion popup

This will ultimately depend on which editor and plugins/extensions you use, so there isn't really a "one size fits all" solution.

xoxorwr:
I'll research on how to do it client side then, thanks!

Issue #711, opened 2025-07-03
label detail broken


bangbangsheshotmedown:
![Image](https://github.com/user-attachments/assets/632c170d-d4b0-4614-ab9f-5270bdbebe0e)

It hurts readability, it should look like:

![Image](https://github.com/user-attachments/assets/f7b45e4c-6798-43e6-b9c9-1c0904bd5a10)

And if you want to keep `proc` then i suggest:

![Image](https://github.com/user-attachments/assets/6c099d4b-0a08-4fdf-ac1e-098bea31e693)

I personally vote for removing `proc`, cleaner

BradLewis:
Yeah I see what you mean. Unfortunately this looked better to me in my nvim configuration, but after checking vscode and the spec, what you've mention here is definitely what's desired.

bangbangsheshotmedown:
Thanks

bangbangsheshotmedown:
By the way, something similar with builtins

![Image](https://github.com/user-attachments/assets/e4c2bcea-81d0-4fce-a00d-a9c0da22a12c)

PR #715, opened 2025-07-03
Fix CompletionLabelDetails for proc groups

BradLewis:
Fixes https://github.com/DanielGavin/ols/issues/711#issuecomment-3032731438


bangbangsheshotmedown:
I was about to submit a PR, i think it's best to convey the information:

image


```
// We have no leading package.Name for the proc
proc_info = proc_info_split[0]
if strings.contains(proc_info, item.label) {
        proc_info = "(..)"
}
```

It's kinda unfortunate that at this point we loose the information of the symbol, perhaps there is a better way of detected exactly if the symbol is a group

BradLewis:
Feel free to create an alternative PR, I just wasn't sure what the appropriate information would be to put there as there isn't a signature for the proc in this case.

Issue #548, opened 2024-11-30
[Feature Request] Show documentation in auto-complete window of the VSCode plugin


flysand7:
When summoning auto-complete menu to chose from one of the procedures I would like to be able to view their documentation without having to insert a procedure call and calling "show hover" command on it. I saw in some other languages' plugins, that vscode has this ability, as shown in the screenshot below:

![image](https://github.com/user-attachments/assets/0badba2c-97cc-46dd-a32f-28f7f3e3f76f)

DanielGavin:
I do return documentation related to the completion item. Not sure how to get vscode to show it.

PR #717, opened 2025-07-04
Feat/add lookup odin root for builtin

BradLewis:
Looks for the `builtin` folder when indexing within `ODIN_ROOT` before looking at the path relative to the executable. Just for the cases where the exe isn't kept at the root of the repo.

I've also added the (..) to match how other languages handle things like proc groups in vscode. It should completely resolve https://github.com/DanielGavin/ols/issues/548 now: 
image


BradLewis:
I'll change the order to look in $ODIN_ROOT only if it doesn't find it in the exe directory so it doesn't cause issues with the CI

PR #738, opened 2025-07-12
Show docs and comments on struct field hover

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/341


BradLewis:
Ahh it looks like there's been a fix for parser with comment groups on the odin side which is causing the issue with the builds. I'll update this PR to remove the parsing we were doing later tonight.

PR #739, opened 2025-07-12
Backport `intrinsics.type_enum_is_contiguous` (https://github.com/odi…

FourteenBrush:
…n-lang/Odin/pull/5455)


BradLewis:
These build errors are just due to some changes on the odin side, should be resolved when https://github.com/DanielGavin/ols/pull/738 gets merged

BradLewis:
This actually fails parsing due to the `where` clause with `Syntax Error: 'where' clauses are not allowed on procedure literals without a defined body (replaced with ---)`. I'll create a PR to remove the `where` clause to match the other procs in this file.

PR #756, opened 2025-07-19
Propagate parameters semantic token

### LINKED ISSUE #679: semantic tokens: variable parameter is not propagated
**### zig**:

Image

**### odin**:

Image

**problem**: self is not underlined

Image

bangbangsheshotmedown:
This is very naive, but this works for me, so far..

This should perhaps be an option tho.. I have 0 clue about lsp

Fix #679


DanielGavin:
I think there is a better way. I'll look at later this week.

DanielGavin:
Fixed it by introducing a new symbol flag for parameters.

Issue #482, opened 2024-08-17
Autocompletions when declaring struct field names


thetarnav:
![image](https://github.com/user-attachments/assets/d814618c-5c8c-4f4e-b6f4-ae0263159c2d)

Completions should probably be disabled when creating new identifiers.

PR #758, opened 2025-07-19
Implement sortText for CompletionItem

bangbangsheshotmedown:
Yoinked the values from zls

Perhaps default could be tuned for Odin, but it looks good to me as is

This should perhaps also be an option, opt in, to not annoy people who are already used to the current non existing "sorting" by label

But i don't know how to do that, so i'll leave it to you

https://github.com/zigtools/zls/blob/45b855f7ec3dccea3c9a95df0b68e27dab842ae4/src/features/completions.zig#L919C23-L920

https://github.com/zigtools/zls/blob/45b855f7ec3dccea3c9a95df0b68e27dab842ae4/src/features/completions.zig#L566-L588

### DIFF
--- a/src/server/completion.odin
+++ b/src/server/completion.odin
@@ -157,11 +157,45 @@ get_completion_list :: proc(
 		get_package_completion(&ast_context, &position_context, &list)
 	}
 
+	for &item in list.items {
+		item.sortText = fmt.tprintf("%s_%s", kind_to_sort_score(item.kind), item.label)
+	}
+
 	if common.config.enable_label_details {
 		format_to_label_details(&list)
 	}
 
 	return list, true
+}
+
+kind_to_sort_score :: proc(kind: CompletionItemKind) -> string {
+	score: string
+	#partial switch kind {
+	case .Module:
+		score = "1"
+	case .Folder:
+		score = "2"
+	case .File:
+		score = "3"
+
+	case .Operator:
+		score = "1"
+	case .Field, .EnumMember:
+		score = "2"
+	case .Method:
+		score = "3"
+	case .Function:
+		score = "4"
+	case .Text, .Constant, .Variable, .Struct, .Enum, .TypeParameter:
+		score = "5"
+	case .Snippet:
+		score = "6"
+	case .Keyword:
+		score = "7"
+	case:
+		score = "8"
+	}
+	return score
 }
 
 get_attribute_completion :: proc(
--- a/src/server/types.odin
+++ b/src/server/types.odin
@@ -369,6 +369,7 @@ CompletionItem :: struct {
 	deprecated:          bool,
 	command:             Maybe(Command),
 	labelDetails:        Maybe(CompletionItemLabelDetails),
+	sortText:            Maybe(string),
 }
 
 CompletionItemLabelDetails :: struct {


BradLewis:
So according the the spec, sorting and filtering is supposed to be handled by the client.

> to achieve consistency across languages and to honor different clients usually the client is responsible for filtering and sorting. This has also the advantage that client can experiment with different filter and sorting models. However servers can enforce different behavior by setting a filterText / sortText

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

I have nothing really against the idea of having a "default" sort for the completions, though I'm not really sure what that should be. And I wouldn't want to be maintaining flags for custom sort orders as different people want different ordering.

DanielGavin:
I don't think we can rely on the client to sort the completions, it doesn't have enough context.

I agree we need some default ordering system. And yes we don't want to maintain a lot of flags. 

If there should be flags, it should be floating point numbers describing the important of something in relation to the scoring.

i.e. `package_importance_modifier: 1.2`, `proc_return_type_matching_proc_param_modifier: 1.5`

So we don't have to do any maintaining and the user can just change the modifiers. We just have a default series of modifiers. And if somebody wants to turn something off, they can just set it to:  0.

BradLewis:
Actually in this case we can rely on the client as the completion response contains the kind of the symbol being suggested. `nvim-cmp` also contains comparators that would do what this PR is trying to do: https://github.com/hrsh7th/nvim-cmp/blob/main/lua/cmp/config/compare.lua#L73 (with the default values for the kinds here https://github.com/hrsh7th/nvim-cmp/blob/main/lua/cmp/types/lsp.lua).

Now I have no idea how one would do something like this in vscode which seems to be what most people are using (is it even possible?). So we could add some custom ordering options as you mentioned which can easily just be added at the final step if it's set.

DanielGavin:
In this case you probably could. But it would have to done for every editor. And in cases of requiring inference context, you would never be able to do it, so it's probably better just to do it all on the server.

There is also some people that wanted the ability to completely remove completion for items that are not possible in the current context: https://github.com/DanielGavin/ols/issues/482

This should of course be behind a enable flag. It could also make the completion more brittle if the parser can't handle partial expression correctly and now completion shows nothing.

BradLewis:
Oh yeah for sure, I just mean in the context of "these are all valid but I would like them to be in x-order" they can (and should) be handled by the client. But if it's us providing things like invalid symbols during struct declarations then we should definitely filter those out.

DanielGavin:
But the client can't do everything:

```
my_int_procedure :: proc() -> int ---

my_procedure :: proc(a: int) ---

main :: proc() {
    my_int: bool
    my_bool: bool

    my_procedure(my*)

}

```

In this case the order should be: `my_int, my_int_procedure,  my_bool`


Evaluation:
* Locals should give higher score
* my_int_procedure actually returns a int, and therefore should have higher score than the boolean variable
* my_bool is the wrong type and should be removed or have lower score.
* my_procedure is not valid

In this case `my_int` and `my_int_procedure` are both valid, but it's more likely that a local variable is going to be used than a global one.

BradLewis:
I think we're discussing 2 different things here. I do agree that in cases such as the above, we can do some scoring to help make the suggestions more relevant. All I am referring to is that in the cases where we have a list of valid entries, a generic "I prefer to have x displayed above y" should be handled by the client as the client already has access to that information and can handle it themselves and it's something that is going to be user dependent anyways.

DanielGavin:
I'm just not sure if it is possible to seperate them:

```
	/**
	 * A string that should be used when comparing this item
	 * with other items. When omitted the label is used
	 * as the sort text for this item.
	 */
	sortText?: string;
```

I don't think the protocol is that great for solving this problem.

Maybe we'll just start with the filtering out invalid symbols, since that would remove a lot of completion pollution. And just let the client handle sorting the completion items.

BradLewis:
Fair enough. It looks like `rust-analyzer` does some analysis like this to provide completion scores that we could do as well. https://github.com/rust-lang/rust-analyzer/blob/master/crates/ide-completion/src/item.rs#L251

But it's worth noting they (`gopls` as well) don't filter out invalid symbols when trying to find completions for function calls. Instead it's just that valid types are scored higher which I think is a good approach.

DanielGavin:
gopls follows this hack, which we also would be forced to do:

```
// This is a hack so that the client sorts completion results in the order
// according to their score. This can be removed upon the resolution of
// https://github.com/Microsoft/language-server-protocol/issues/348.
SortText: fmt.Sprintf("%05d", i),
```

Which I then assume rustanalyzer also does, which brings us back to the server doing all the ordering.

The lsp issue was closed and was decided the client should decide, not the server.

Which brings us back to not allowing for scoring and only filtering out invalid symbols.

DanielGavin:
Closing it. There is a plan to add support for improving the completion order using SortText. But it's going to be a bit different.

Issue #330, opened 2024-03-29
Display calling convention on hover


thetarnav:
The calling convention of procedures is part of it's type, it would be nice to display it on hover, if it's something else than the default `"odin"`.

Similarly attributes and directives would be useful as well, especially `#optional_ok`, `#force_inline`, `@(require_results)`, etc.

![image](https://github.com/DanielGavin/ols/assets/24491503/308f42d8-3702-43d3-a1f2-3111315dfa8a)

PR #762, opened 2025-07-20
Add proc directives, calling convention and attributes to the hover information

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/330
image

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -1038,7 +1038,7 @@ internal_resolve_type_expression :: proc(ast_context: ^AstContext, node: ^ast.Ex
 	case ^Map_Type:
 		return make_symbol_map_from_ast(ast_context, v^, ast_context.field_name), true
 	case ^Proc_Type:
-		return make_symbol_procedure_from_ast(ast_context, node, v^, ast_context.field_name, {}, true), true
+		return make_symbol_procedure_from_ast(ast_context, node, v^, ast_context.field_name, {}, true, .None), true
 	case ^Bit_Field_Type:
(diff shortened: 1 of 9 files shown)


BradLewis:
I think this will impact the completions in vscode in unintended ways, I'll have a look over that to make sure it's not too polluting.

Edit: For now I've only enabled this on hover information. Might need to look into reworking how the completion information is constructed when using `labelDetails`.

DanielGavin:
Yeah it's a bit of problem when it comes to adding more detail. You really need to check different editors, since you really never know how it will look.

I'll try and check it on vscode before merging it.

DanielGavin:
It could also be interesting to experiment with markdown in the completion items for documentation, since the hover uses them aswell:

```
/**
	 * A human-readable string that represents a doc-comment.
	 */
	documentation?: string | MarkupContent
```

The completion documentation looks worse compared to the hover one.

BradLewis:
Yeah I noticed that as well, the `detail` renders nicely in neovim but not at all in vscode. Might want to move those details to the docs instead so that we can use markdown for it.

BradLewis:
So I had a play around with adding markdown documentation to the completions and it made quite the difference in vscode as all the text is render nicely and distinctly, to the point that maybe we didn't need to move the comments after all lol (though I do thinks it's better this way, easier in the code too).

Without markdown in vscode: 
image

With: 
image

I think I'll start reworking the completion code to work with this markdown rather than the plaintext details. Should hopefully help with the scoring as well when we move to do that.

DanielGavin:
Sounds good moving them to documentation, probably also more correct in terms of following the protocol.

PR #767, opened 2025-07-20
Log version when starting ols

BradLewis:
To help with common debugging requests, it would be good for us to simply print the version so we can tell if a user is using an up to date version of ols. It seems to be a common issue where people report issues on vscode, and have their vscode extension up to date but not necessarily the ols binary, and it would be good to be able to verify what version they're on easily.


DanielGavin:
Might also be good to make nightly actually nightly. Right now I do it manually, sometimes I forget to build it after big changes.

BradLewis:
Yeah that's a good idea. It also looks like this most recently nightly didn't include this change so it would be good to help with issues like that 😅

PR #770, opened 2025-07-21
Label Details description should show type information only

xoxorwr:
It was too much noise it became unusable

Before:

image

After:

image

The detail box should perhaps get rid of anything before `//`, but I can't see where the logic is

### DIFF
--- a/src/server/completion.odin
+++ b/src/server/completion.odin
@@ -2150,9 +2150,16 @@ format_to_label_details :: proc(list: ^CompletionList) {
 			}
 		case .Variable, .Constant, .Field:
 			type_index := strings.index(item.detail, ":")
+			type_name := item.detail[type_index + 1:]
+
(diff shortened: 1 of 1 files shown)


src/server/completion.odin
```diff
@@ -2150,9 +2150,16 @@ format_to_label_details :: proc(list: ^CompletionList) {
...
+
+			commentIndex := strings.index(type_name, "/")
+			if commentIndex > 0 {
+				type_name, _ = strings.substring(type_name, 0, commentIndex)
+			}
+
 			item.labelDetails = CompletionItemLabelDetails {
 				detail      = "",
-				description = item.detail[type_index + 1:],
+				description = type_name,
```
BradLewis, line 2162: Looks good to me. We should also probably remove the comments from the proc details as well.

Issue #773, opened 2025-07-22
No completion for bitsets when calling proc with named argument


bangbangsheshotmedown:
Example:

```odin

Test :: enum {
    A, B, C
}

Flags :: bit_set[Test]


testme::proc(skip: int = 0, flags: Flags = {})
{
}


main :: proc() {
                   ┌ // no completion
    testme(flags={.|})

                ┌ // works
    testme(0, {.|})
}
```

bangbangsheshotmedown:
Thanks

PR #775, opened 2025-07-22
Correctly complete bitsets with named proc args

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/773

### DIFF
--- a/src/server/completion.odin
+++ b/src/server/completion.odin
@@ -1177,9 +1177,21 @@ get_implicit_completion :: proc(
 					if !arg_type_ok {
 						return
 					}
+					if position_context.field_value != nil {
+						// we are using a named param so we want to ensure we use that type and not the
+						// type at the index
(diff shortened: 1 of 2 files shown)


src/server/completion.odin
```diff
@@ -1177,9 +1177,22 @@ get_implicit_completion :: proc(
...
+						if name, ok := position_context.field_value.field.derived.(^ast.Ident); ok {
+							if i, ok := get_field_list_name_index(name.name, proc_value.arg_types); ok {
+								arg_type = proc_value.arg_types[i]
+							}
+						}
+					}
 
 					if enum_value, ok := unwrap_enum(ast_context, arg_type.type); ok {
 						for name in enum_value.names {
+							log.info(position_context.comp_lit)
```
DanielGavin, line 1192: Forgot to remove debug info

BradLewis: Good catch! Should be good now.

PR #776, opened 2025-07-23
Add version information to the build scripts

BradLewis:
Also corrects fetching the commit and date in windows.


DanielGavin:
Works fine on Windows.

BradLewis:
Yeah I've actually properly set up a dev environment on windows so I can more cleanly test these kinds of things

PR #777, opened 2025-07-24
Check if a value declaration is a comp lit for semantic tokens

BradLewis:
After a discussion on discord, it looks like the semantic tokens doesn't distinguish between type declarations and constant values. This is a first pass at checking if a value of a `Value_Decl` is a `Comp_Lit` or not. I'm not sure how many possible combinations there are here so this is either overkill or going to be missing some cases.


DanielGavin:
Tried different cases with this branch, and it seems fine to me.

BradLewis:
Okay perfect, I'll undraft and merge it in then.

Issue #779, opened 2025-07-24
builtin function causes variable to be missing from completion list


bangbangsheshotmedown:
example:

```odin

main::proc() {
    checkme := max(f32)

    check // no completion
}

```

PR #780, opened 2025-07-25
Resolve builtin max, min, clamp and abs return types correctly

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/779

Unfortunately there's a bit of massaging that needs to happen to resolve these functions as they're pretty non standard, and the actual returned value isn't determined until it hits the checker. This essentially just tries to replicate the basic behaviour. There might be some edge cases that I have wrong, but most of the time it seemed to follow this, or it would be an error in the checker so we don't need to concern ourselves that much.

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -933,6 +933,157 @@ resolve_basic_directive :: proc(
 	return {}, false
 }
 
+// Gets the return type of the proc.
+// Requires the underlying call expression to handle some builtin procs
+get_proc_return_types :: proc(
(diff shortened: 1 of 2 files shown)


DanielGavin:
Yeah we are bound to custom handle many of the builtin procedures that aren't real Odin code. Yeah not having to validate the code and letting the checker do it, simplifies ols a lot.

We haven't even started looking at the where clauses for proc groups. That will be a bit of work.

example: 
```
where intrinsics.type_is_numeric(E) 
```

Though no one has ever complained about this yet.

Issue #787, opened 2025-07-27
Crash on `dxgi.IFactory6` when `enable_semantic_tokens: true`


denovodavid:
Getting a crash with the following when `enable_semantic_tokens` is `true`:
```odin
package foo_bar

import "vendor:directx/dxgi"

// working
foo: ^dxgi.IFactory5

// crashes
bar: ^dxgi.IFactory6
```

Language server output:
```
[Error - 10:03:25 AM] Starting Odin Language Server nightly-2025-07-26-b9a91fe
D:/a/ols/ols/Odin/base/runtime/dynamic_map_internal.odin(427:4) panic: unable to insert into a map
[Error - 10:03:25 AM] The Odin Language Server Client server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.
[Error - 10:03:25 AM] Server process exited with code 3221225501.
```

System info:
```
OS:      Windows 11 Professional (version: 24H2), build 26100.4652
CPU:     AMD Ryzen 9 9950X 16-Core Processor
RAM:     31911 MiB
```

BradLewis:
Well this is a weird issue. It seems to fail to find a slot to insert the entry into as you can see here: https://github.com/odin-lang/Odin/blob/master/base/runtime/dynamic_map_internal.odin#L427. I have no idea why this would only happen with semantic tokens too when it seems I can hover over `IFactory6` just fine.

I've managed to make it go away by setting a small capacity for the map. I'll push that quickly so it stops crashing for you.

PR #788, opened 2025-07-27
Increase the capacity of the `usings` map so it doesn't cause a panic when trying to insert

BradLewis:
No idea why this is an issue, but this seems to resolve it. It fails to find a free slot to insert an item into as you can see here https://github.com/odin-lang/Odin/blob/master/base/runtime/dynamic_map_internal.odin#L427. Only happened with semantic tokens, hover and everything else was fine.

Resolves https://github.com/DanielGavin/ols/issues/787


DanielGavin:
Yeah that's strange and concerning.

PR #792, opened 2025-07-28
Improve completion processing

BradLewis:
First pass at the improvements to completions. It's based around providing the symbols rather than the complete `CompletionItem` and having a central place that fills out all the relevant information. Doing this made it easy to consolidate the documentation so now it should contain very similar information that hover uses, and it will use markdown rather than plaintext which is especially apparent in things like vscode:

image

It also optionally allows you to directly provide the completion item as there's a lot of code that does that right now, and I imagine there will always be cases that don't correspond to a symbol.

With this we should be able to do some analysis on the symbols to score them for sorting.

I've done some basic testing on vscode as well as my general neovim usage. Should probably be tested more just to be sure I didn't miss anything.

On another note these changes also made some issues with flakiness in the tests much more apparent (it would fail 100% of the time rather than randomly on occasion). Seems like we were just missing some `@(thread_locals)`.

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -891,7 +891,7 @@ resolve_basic_lit :: proc(ast_context: ^AstContext, basic_lit: ast.Basic_Lit) ->
 	out commented because of an infinite loop in parse_f64
 	else if v, ok := strconv.parse_f64(basic_lit.tok.text); ok {
 		value.type = .Float
-	} 
+	}
 	*/
 
 	symbol.pkg = ast_context.current_package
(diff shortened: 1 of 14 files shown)


DanielGavin:
Makes good sense to move the completion documentation to a single place.

DanielGavin:
Seems good to me. I wonder if we should consider having a nightly version and a release version. So it's more of an opt-in thing for the newest changes.

BradLewis:
We could also potentially have an automated weekly nightly release to reduce the amount of updates and leave the manually option for releasing urgent hotfixes. I think the biggest problem with having multiple releases would be updating the vscode extension to support them lol

Issue #488, opened 2024-08-24
[Feature Request] Struct size and alignment and padding for fields on mouse hover


dgoalcantara:
Qt Creator has those really nice features. It can also show those values for Union., size in bytes for enum values or most variable/types.

Not sure if possible or worth the effort. Qt Creator uses Clang for it's code model.

Example:
![image](https://github.com/user-attachments/assets/1ce5c378-1ab4-4912-9134-1da99b3ee360)

![image](https://github.com/user-attachments/assets/8e0dbcbd-a4f7-4a3d-96cb-f0c7e1922d73)

Sorry for the image quality: took those from my phone.

PR #796, opened 2025-07-28
WIP struct size info on hover

xdk78:
Very rough draft version of hover struct size info.
Resolves: #488 
image

todo:
- [ ] calculate size, padding, and alignment for each field
- [ ] calculate for more complex types including arrays, matrices, etc.


Bambo-Borris:
Nice to see some activity towards this! What is left outstanding for this to be ready to review?

Issue #799, opened 2025-07-29
Helix reporting "language server exited" during completion


Bambo-Borris:
Noticing helix report that ols has exited at random points when doing completion. It seems kind of random when/where it will occur, but once it does I basically have to restart the LSP and that can mean 2-3 restarts if it's playing up. Hadn't had issues with this so I did a bisect and it seems the issue came in after #774.

I attached a debugger to ols with a debug build, didn't get anything interesting. I will note I do have "enable_references" and "enable_fake_methods" set in my config. Besides that there's nothing perculiar about the setup. 

I tried reproducing the issue in a single file project with no success also. If I had a good test source example I'd provide it but I might just have to do some extra debugging and feedback to this if I find anything to help identify the cause of the issue.

BradLewis:
DId you get information about where it was crashing from the debugger?

PR #800, opened 2025-07-30
Fix issue with constructing docs on indexed symbols

BradLewis:
Whilst looking into https://github.com/DanielGavin/ols/issues/799 I noticed that I was getting null bytes for symbol method documentation after first completion. After a brief investigation, I found that with the recent changes it was overriding the indexed symbols documentation with a string created on the temp allocator. To fix this for now I've just made it so that it no longer overrides the docs in this case.

A more permanent solution to this will come with the ongoing improvements to the docs generation so that we always use those functions to construct the docs as needed. With that becoming more streamlined it should make it easier as well.

Possibly resolves https://github.com/DanielGavin/ols/issues/799


BradLewis:
Looks like there's an issue with the build after https://github.com/odin-lang/Odin/pull/4140 was merged. I'm not sure why it would run the `when ODIN_ARCH == .arm64` when this is an x86_64 machine?

PR #807, opened 2025-07-30
Improve builtin proc resolution, complex and quaternion procs and add missing basic type keywords

BradLewis:


DanielGavin:
Looks great. Good idea to move the builtin procedures to a new file.

There is probably a lot more to move from analysis. It's getting pretty big. Maybe all the local gathering related procedures could be moved to it's own file.

BradLewis:
I was actually thinking the same thing precisely for the locals code as I always seem to struggle to find it quickly lol

PR #817, opened 2025-08-01
Fix/resolve procs that return procs called immediately

BradLewis:
Improves resolution of symbols when immediately calling a proc returned from a proc (eg `a := foo()()`).

In a draft at the moment as I'm not sure it's complete yet, but I'm running into to issues with some tests crashing (`objc_return_type_with_selector_expression_2` and `ast_generic_struct_with_array`) on macos (running them on windows is fine) due to the code around `1199` in `analysis.odin`. The broken tests don't go through the if statement on line `1196` so I'm not sure what's causing them to fail. 

Looks like all the tests pass here 🤔 

Edit: It looks like moving the logic from inside the switch statement to external procs has pretty significantly reduced the amount of stack being used (from ~50kb per function call to ~34kb). Definitely something to keep an eye on when we're dealing with stack overflows

Issue #578, opened 2025-01-10
Ols presumably crashes on large malformed files


SaculRennorb:
I'm porting a project from cpp to odin, which contains files that are several thousand files long. My process for doing so is regex replacing parts of the cpp code into the odin semantics. 

I was using ols (vscode) for the verry first (small) file, but after opening one of the larger ones it just silently crashes and subsequently fails to restart. 

Ive tried the stable version, the nightly version and even built ols from master a9853fc, none of it changed anything. I also don't get any debug output (afaict), not even using the debug build of ols. 

Here is the output from vscode:

```
[Info  - 5:42:34 AM] Connection to server got closed. Server will restart.
true
[Info  - 5:42:35 AM] Connection to server got closed. Server will restart.
true
[Info  - 5:42:35 AM] Connection to server got closed. Server will restart.
true
[Info  - 5:42:35 AM] Connection to server got closed. Server will restart.
true
[Info  - 5:42:35 AM] Connection to server got closed. Server will restart.
true
[Error - 5:42:35 AM] The Odin Language Server Client server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.
```

Its not reasonable to attach the source file (because of the length), but if you can tell me how to get more debug output / logs i can try to provide that. 

Since it presumably immediately crashes i also failed to attach a debugger, and this is where i just submit this issue as is.

DanielGavin:
You can open up an odin file that doesn't crash. Attach the debugger. Open the file that crashes. I doubt any log would be useful. 

You might be able to upload the source code file without printing it.

SaculRennorb:
> You can open up an odin file that doesn't crash. Attach the debugger. Open the file that crashes
good call, as you can tell form the timestamps it was quite early for me. 

It appears to be a stack overflow in the odin tokenizer. I'm guessing that is outside your direct control, but maybe you can at least figure out some more details. 

Here you also have the broken file, maybe it helps in some way: 
[imgui.zip](https://github.com/user-attachments/files/18377026/imgui.zip)

BradLewis:
I was able to reproduce this as well. It crashes when parsing within the odin parser due to a stackoverflow. Rebuilding `ols` with a larger stack did resolve the issue for me (16mb failed but 32mb was fine).

BradLewis:
Should be fixed now with the recent improvements made to the stack usage.

PR #818, opened 2025-08-02
Switch to using out params for `internal_resolve_type_expression` to greatly reduce stack used

BradLewis:
After working on https://github.com/DanielGavin/ols/pull/817 and encountering some issues with certain tests stack overflowing which led to a discussion here https://github.com/odin-lang/Odin/issues/5528, it looks like having large switch statements completely blows out the stack. The worst culprit for us right now is the `internal_resolve_type_expression` function which was adding around 50kb per call (~25kb with `-o:speed`). By using an out parameter for just this function it pretty significantly reduces the amount of stack used (from ~500kb to ~240kb in the tests that were stack overflowing for me).

I could probably add more functions to behave like this, but since this is the main culprit I think we can just make this change here to minimise the impact, as out params aren't exactly the nicest.

I also retested this https://github.com/DanielGavin/ols/issues/578 and with this change (and the other ones in the previous PR) this large file no longer stackoverflows for me.

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -695,7 +695,8 @@ resolve_function_overload :: proc(ast_context: ^AstContext, group: ast.Proc_Grou
 	candidates := make([dynamic]Candidate, context.temp_allocator)
 
 	for arg_expr in group.args {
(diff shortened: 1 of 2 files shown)


DanielGavin:
Yeah not the nicest looking way of doing things, but it's worth it for the stack reduction.

Issue #825, opened 2025-08-05
Extension can become permanently broken if user dismisses "update language server" popup


jack-jjm:
VSCodium 1.100.33714
Linux Mint

I'm in an odd kind of difficulty here and I decided I'd just report the whole scenario, even though some of it is to do with VSCode more than this extension.

Since I installed the OLS addon on this computer, I've found that every time I open VSCode, I get a popup prompting me to install the Nightly update to the language server. This is annoying, so today I hit the little cog icon on that popup and selected something that said "do not show addon notifications" or something like that to get rid of it for good.

But now, my OLS addon doesn't work at all. I see this in the logs:

```
INFO [8/5/2025, 11:56:41 AM]: PersistentState: {
  lastCheck: 1754242321673,
  releaseId: 237172629,
  serverVersion: '0.1.38'
}
DEBUG [8/5/2025, 11:56:41 AM]: Issuing request for released artifacts metadata to https://api.github.com/repos/DanielGavin/ols/releases/tags/nightly
```

But other than that it's as if the extension isn't even there.

So I guess there are three points here:

1. Not having the language server updated to the latest version, or choosing not to download it, should not break the addon.
2. There should be an addon setting to disable checking for language server updates, or to change from the nightly build to something less frequently updated.
3. In the meantime, I need to figure out how to get that damn popup to appear again so I can install the language server and get my addon working again (OK I figured this out, you hit the bell in the bottom right).

BradLewis:
I've created a PR that should fix the issue with the extension just hanging when you hide the notification. I think we could also make "nightly" just be a weekly release to stop the constant prompting for updates, and allow people who want the more frequent updates to just build off of master.

PR #826, opened 2025-08-05
Check for updates in the vscode extension asynchronously

BradLewis:
Basically rather than checking for updates inline when starting the extension, we just check it without blocking the start of the lsp.

The problem was when people would hide the notification, it would sit waiting for them to respond before moving forward. This should mean that it'll always run and only update if they confirm.

We could also implement a setting now that would allow people to specify how often they want to check for the nightly updates since some people seem to be annoyed by the popup every day.

I did some basic testing on my end and it seemed good but should have another person try it out. I tested with/without updates available, and also a completely fresh install and it seemed to work fine.

Helps resolve https://github.com/DanielGavin/ols/issues/825

### DIFF
--- a/editors/vscode/src/extension.ts
+++ b/editors/vscode/src/extension.ts
@@ -61,6 +61,8 @@ export async function activate(context: vscode.ExtensionContext) {
 		throw new Error(message);
 	});
 
+	checkForUpdates(config, state, false)
+
 
 	const workspaceFolder = vscode.workspace.workspaceFolders?.[0];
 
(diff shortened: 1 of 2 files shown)


editors/vscode/src/extension.ts
```diff
@@ -333,35 +359,30 @@ async function getServer(config: Config, state: PersistentState): Promise<string
...
+async function checkForUpdates(config: Config, state: PersistentState, required: boolean) {
+	const platform = getPlatform()
+	const release = await downloadWithRetryDialog(state, required, async () => {
 		return await fetchRelease("nightly", state.githubToken, config.httpProxy);
 	});
 
 	if (release === undefined || release.id === state.releaseId) {
 		await state.updateLastCheck(Date.now());
-		return destExecutable;
+		return "";
```
DanielGavin, line 385: Surprised typescript allows you to return ""; when the signature is void.

DanielGavin:
Strange. It still doesn't seem to work for me. I tried installing the extension from vsix after building it. If i don't press the update, it still doesn't work.

BradLewis:
That's weird. On my end it works fine, I just reinstalled it to try again. You can see the log showing that it's starting, I get hover information and it's also prompting me to update the binary:
image

I did a test on windows as well and got the same result

DanielGavin:
I think mine was caused by: `const serverPath = await bootstrap(config, state)`

This checks if there is a ols binary at all, but it uses the extension id in it's path. So if you use a new extension with an existing ols binary downloaded from an earlier version, it can't find it.

`boostrap` basically uses `checkForUpdates`, but it awaits it, so you have to prompt yes before ols activates.

BradLewis:
Yeah if it can't find the binary you'll need to respond to the prompt as it doesn't have anything it can run.

PR #843, opened 2025-08-10
Fix issues initialising globals due to odin @init changes

BradLewis:


BradLewis:
Ah I see this has already been fixed, nevermind then!

PR #876, opened 2025-08-13
Fix crash on accessing nil arg type pointer

thetarnav:
Fixes this:
```odin
import "core:encoding/json"

main :: proc () {

	src: []byte
	data: struct {}

	json.unmarshal(src, &data, /* type .JSON5 */, context.allocator)
}
```


BradLewis:
Thanks!

PR #972, opened 2025-09-06
Only inline single line switch if line is not too long

BradLewis:
I'm sure there's a better way to handle this but I wasn't able to figure it out. Essentially just want it to optionally insert a newline if the resulting line is too long. Currently enabling this means that things like long proc calls won't be wrapped across multiple lines as you would expect; it just adds the whole thing to the one line even if it's 200 characters long.

This is just an easy check to only go down the single line path the resulting line would fit within the character limit. This definitely feels like a hack though as none of the other parts of the code require knowing the character width, that's just something that's used when handling the printing.

Let me know if you know of the proper way to do this.

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -1086,7 +1086,7 @@ visit_stmt :: proc(
 		if count := len(v.body); count > 0 {
 			set_source_position(p, v.body[0].pos)
 			if count == 1 && p.config.inline_single_stmt_case {
-				document = cons_with_opl(document, nest(visit_stmt(p, v.body[0])))
+				document = group(nest(cons_with_opl(document, nest(visit_stmt(p, v.body[0])))))
 			} else {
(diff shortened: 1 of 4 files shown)


DanielGavin:
I would be careful with adding too many config flags, it might become quite annoying to maintain in the formatter.

You can call a `group` function. That should be enough to do what you want.

```
document = group(cons_with_opl(document, nest(visit_stmt(p, v.body[0]))))
```

The way it works is that all optional newlines get evaluated as a group. So if you had another `opl` that would also break at the same time if the character width is exceeded for the entire group.

This document system works quite a lot like `prettier` and `rustfmt`.

BradLewis:
Yeah I'm pretty hesitant to add more configuration especially for the formatter. This was just something that was already added but wasn't behaving as expected with these long lines.

That indeed did the trick. I'm surprised that I somehow didn't try that in all my testing. But thanks! I'll make the correction.

PR #975, opened 2025-09-07
First pass at scoring completion items

BradLewis:
Applies a very basic scoring system to the completion items. Also sorts by length when scores are the same which is similar to how other LSPs behave.


DanielGavin:
Seems like a good start for scoring.

BradLewis:
Yeah it's very basic but it does seem to make a nice difference with my neovim setup.

DanielGavin:
Yeah it seems better for me as well. I guess we'll soon hear people complaining about it or praising it :)

Issue #978, opened 2025-09-07
[Bug] `When` blocks with parentheses aren't evaluated.


BigBoyBarney:
Hi! Thank you for your work on Ols!

Currently, when blocks with parentheses around the condition aren't evaluated, and subsequently do not appear in autocomplete.
Here's a video and example code illustrating the issue:
```odin
package test

// Autocomplete does not see this if the parentheses are present.
// If you remove the parentheses, they show up correctly.
when (ODIN_OS == .Linux) { 
	Test_Enum :: enum {
		One,
		Two,
		Three,
	}

	Test_Struct :: struct {}
}

enum_test :: proc(test: Test_Enum) {  }

main :: proc() {
	enum_test()
}
```
Video example:
https://files.catbox.moe/05dh5j.mp4

PR #979, opened 2025-09-07
Correctly resolve when conditions wrapped in parens

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/978


BigBoyBarney:
You are incredible, thank you so much

PR #982, opened 2025-09-07
Remove completions when entering bit_field or struct field names

BradLewis:
Removes all completions when entering names for `bit_field` or `struct` types. I don't think there's ever a case where it's actually valid to provide a symbol as a struct name? So we just remove all completions in this case, which is also how `rust-analyzer` behaves.

Resolves https://github.com/DanielGavin/ols/issues/482


DanielGavin:
Yeah don't know any case where it would be needed.

PR #987, opened 2025-09-09
Cache workspace pkgs so we don't walk the file system each workspace symbol request

BradLewis:
It was mentioned on the discord that the workspace symbols can be slow for large codebases. After a brief look into it I noticed we were walking the filesystem to find the files to check. This is fine as a one off but since it makes the request many times when you filter the symbols (basically every time you pause typing) this adds up and is noticeable. 

With this we basically just cache the packages with a 1 minute TTL so we don't continuously do this walk. On the Odin repo on my m1 pro it took the average response time from 50ms on the low end to up to 300ms on the high end, to a very consistent 40ms or so on average.

### DIFF
--- a/src/server/indexer.odin
+++ b/src/server/indexer.odin
@@ -73,19 +73,8 @@ fuzzy_search :: proc(
 	bool,
 ) {
 	results, ok := memory_index_fuzzy_search(&indexer.index, name, pkgs, current_file, resolve_fields)
-	result := make([dynamic]FuzzyResult, context.temp_allocator)
-
 	if !ok {
 		return {}, false
 	}
-
-	for r in results {
-		append(&result, r)
-	}
-
-	slice.sort_by(result[:], proc(i, j: FuzzyResult) -> bool {
-		return j.score < i.score
-	})
-
-	return result[:], true
+	return results[:], true
 }
--- a/src/server/workspace_symbols.odin
+++ b/src/server/workspace_symbols.odin
@@ -1,15 +1,23 @@
 package server
 
-
 import "core:fmt"
 import "core:log"
 import "core:os"
 import "core:path/filepath"
 import "core:strings"
+import "core:time"
 
 import "src:common"
 
 dir_blacklist :: []string{"node_modules", ".git"}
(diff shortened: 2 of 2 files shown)


DanielGavin:
Yeah seems fine to make a cache. Should probably have a data structure of all the files we have, and eventually what dependencies each file has. 

Yeah the `slice.sort_by` makes no sense.

BradLewis:
It would be nice if we had a more reactive way to determine this but right now it seems we don't receive any information when a new folder is added to the project. 

Edit: actually I was playing around with this and at least with neovim we get notified whenever we add any files to the project in or out of neovim. However when I was testing in sublime, it did not notify us of any change. Seems like it's just not supported? https://github.com/sublimelsp/LSP/issues/892

DanielGavin:
Yeah it's not supported on all editors.

It also is registered quite differently with dynamic registration.

DanielGavin:
@BradLewis In the same issue it mentioned a plugin for sublime that adds the notification: https://packagecontrol.io/packages/LSP-file-watcher-chokidar

I already use the notification to build package alias for collection every time a file gets created or deleted.

`clear_all_package_aliases`
`find_all_package_aliases`

BradLewis:
Yeah I noticed that as well so at least people would be able to patch it to work. 

Right now sublime will tells us when the files are changed, just not when they are added or deleted. I was thinking of using these messages to maintain the list of packages in the workspace, but right now I'm unsure if sublime (or other editors for that matter) actually tell us that they won't send the create/delete updates, so I don't know if we'd be able to distinguish between "I haven't received any notifications because nothings happened" vs "I haven't received any notifications because the client isn't telling me".

If we can make that distinction though it would be good to just use it to maintain the list so we can just walk to filesystem once on start to map the packages and then update them when we get the messages, and then if we have a client that won't tell us we can just use this basic time based one. 

This should also help with the performance of references and renaming as it also walks the filesystem to find the necessary packages to check, though this time based cache probably wouldn't be useful unless you're doing a bunch of renamings back to back or something.

Issue #989, opened 2025-09-10
[Feature request] Display `where` clauses in parapoly proc hover


BigBoyBarney:
Hi! Thank you for your work on Ols!

Currently, `where` clauses in parapoly procedure definitions are not reflected in the type hint popup / autocomplete information.

<img width="874" height="249" alt="Image" src="https://github.com/user-attachments/assets/8f6a7f54-241c-4214-a35d-1d0a3057245f" />

This was mentioned on Discord as well, but I thought I'd open an issue for posterity :D

PR #990, opened 2025-09-10
Add where clauses to hover information

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/989


BigBoyBarney:
Thank you!!

PR #996, opened 2025-09-11
Inlay hints fixes

### LINKED ISSUE #901: Inlay Hints Incorrect When Using Multi-return `proc` In Another `proc` Which Accepts More Parameters
I would expect the first hint to look something like `one, two =`, and the last to be `three =`:
Image
```odin
package main

main :: proc() { takes_three(returns_two(), 3) }
takes_three :: proc(one, two, three: int) {}
returns_two :: proc() -> (int, int) { return 1, 2 }
```

This is another example which appears differently by using a default parameter, where I would only expect two total—`one, two =`, and a `three =`:
Image
```odin
package main

main :: proc() { takes_three(returns_two(), 3) }
takes_three :: proc(one, two: int, three := 3) {}
returns_two :: proc() -> (int, int) { return 1, 2 }
```
---
ols:     dev-2025-08-18-6282669
Odin:    dev-2025-08:aff96e7d3
OS:      Windows 11 Professional (version: 23H2), build 22631.5624
CPU:     AMD Ryzen 7 7700 8-Core Processor
RAM:     61033 MiB
Backend: LLVM 20.1.0

### LINKED ISSUE #920: Wrong inlay-hint with named arg after var-args
```odin
fmt.panicf("Error in %s: %s", operation, error_msg, loc=loc/* ", loc = #caller_location" */)
```

thetarnav:
- fixes #920
- fixes #901
- Makes expected hints to be provided in source code, eg: `my_function([[required = ]]42[[, optional = false]])`
- Improves error messages when testing inlay hints
- adds `build_test` build script command (eg for debugging tests)
- ignores `.o` and `.bin` linux build artifacts

### DIFF
--- a/src/server/inlay_hints.odin
+++ b/src/server/inlay_hints.odin
@@ -38,10 +38,7 @@ get_inlay_hints :: proc(
 
 			if call, is_call := node.derived.(^ast.Call_Expr); is_call {
 				data := (^Visitor_Data)(visitor.data)
-				if !visit_call(call, data) {
-					log.errorf("Failed to get inlay hints for call expression: `%s`",
-						get_ast_node_string(call, string(data.document.text)))
-				}
+				visit_call(call, data)
 			}
 
 			return visitor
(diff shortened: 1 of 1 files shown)


BradLewis:
I tested this a bit but I seem to get a lot of errors about it failing to get inlay hints. For example, in `completions.odin`: 
image

and in `analysis.odin`: 
image

I think these are all overloaded procs so I'm guessing it's just failing with them.

thetarnav:
Probably most come from these lines:
```odin
symbol_and_node := data.symbols[uintptr(call.expr)] or_return
proc_symbol := symbol_and_node.symbol.value.(SymbolProcedureValue) or_return
```
I could change it to just return `true` to prevent the error message.

But this also means that there is a bug with resolving proc groups—I'm not getting any hover info for `append`.

Might as well add better errors.

BradLewis:
It seems to be failing at `symbol_and_node := data.symbols[uintptr(call.expr)] or_return` which is interesting. The problem now is it absolutely spams the logs with these errors as `inlay_hints` seems to be called every time you scroll the code in the editor.

BradLewis:
Looks like there's just issues with it failing to resolve certain proc overloads. It doesn't happen with all of them, but enough that these inlay hints highlight the problems for us.

In any case this is unrelated to your changes, so LGTM, thanks!

Issue #1002, opened 2025-09-13
Auto complete not working inside selector call expression


Dimension4:
Autocomplete for fields inside selector call expressions doesn't seem to work. Here's an example:

```odin
package dumpster

import "core:fmt"

Data :: struct {
	x, y: int,
}

IFoo :: struct {
	bar: proc(self: IFoo, x: int),
}

print :: proc(self: IFoo, x: int) {
	fmt.println(x)
}

main :: proc() {
	data := Data{3, 4}
	foo := IFoo {
		bar = print,
	}

	foo->bar(data.x) // no autocomplete for 'x'
}
```

It seems like `data.` is ignored when trying to resolve the scope, because I get autocomplete suggestions as if I just wrote `foo->bar(`. Here's a screenshot of what I mean:

<img width="980" height="977" alt="Image" src="https://github.com/user-attachments/assets/facdc1e9-0f59-4ae8-9811-fe697894b713" />

I am on fairly recent master branches of odin and ols (within the last week)

BradLewis:
I just realised this didn't actually fix the issue, I'll reopen.

BradLewis:
I'm going to reopen this again as I'm not confident that those changes won't break something else. I'll try to have another look into it this weekend.

BradLewis:
This should finally be working now: 

<img width="471" height="399" alt="Image" src="https://github.com/user-attachments/assets/a0ca643f-38c5-46f3-9635-1b4e5ae95010" />

You will need to make sure you're building `ols` with the latest odin master as the fix required changes in the odin parser.

Dimension4:
Can confirm it works nicely, thanks a lot for fixing!

PR #1011, opened 2025-09-14
Correct selector completions within selector call exprs

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/1002, for real this time.


BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/1002, for real this time.

PR #1015, opened 2025-09-14
Use core:flags instead of custom flag parsing code and fix 2 bugs

godalming123:
- Fixes formatting not working for blank files
- Fixes odinfmt not detecting the config file when `-stdin` is specified because it thinks that it is formatting the `-stdin` path
- Replaces the custom arg parsing code with odin's `core:flags` library


BradLewis:
Thanks for this! I was just doing some testing of your branch and it seems the default behaviour when running `odinfmt` has changed. 

Before it would do:

```sh
❯ ./odinfmt
usage: odinfmt [-w {filepath}] [-stdin]
```

Whereas now it will format all files in the directory and dump the formatted files to stdout.

godalming123:
> Thanks for this! I was just doing some testing of your branch and it seems the default behaviour when running `odinfmt` has changed. 
> 
> Before it would do:
> 
> ```sh
> ❯ ./odinfmt
> usage: odinfmt [-w {filepath}] [-stdin]
> ```
> 
> Whereas now it will format all files in the directory and dump the formatted files to stdout.

IMO, the behaviour in this PR is better than the current behavior. I think that how odinfmt currently works is inconsistent because running `odinfmt -w` does not require the path to be specified, but running just `odinfmt` does. I can change the behaviour in this PR back to the current behavior though if you would like.

BradLewis:
Yeah I think that would be good, just to keep the behaviour between the 2 implementations consistent.

BradLewis:
LGTM thanks!

Issue #917, opened 2025-08-20
Include the value in constants hover info


thetarnav:
Hover info for constants should include the value where it makes sense,
instead of always trying to just show the type.

```odin
A :: 123    // "A: int"   --  should be "A :: 123"    -- FOO could be used as a float too

B: u8 : 123 // "B: u8"    --  should be "B: u8 : 123" -- It's not obvious from "B: u8" that B is even a constant

C :: 1.23   // "C: float" --  should be "C :: 1.23"   -- float is not a type
```

This might be tricky with expressions, as I don't think ols can resolve these yet.
```odin
A :: B + C
```
But displaying `A :: B + C` would be fine too,
or `A: u8 : B + C` if the type can be inferred (https://github.com/DanielGavin/ols/issues/916)

This would also allow to correct the hover info for types:
```odin
Struct :: struct {a, b, c: int}
// hover info:
//     Struct: struct {   // <--- should be "Struct :: struct {"
//         a: int,
//         b: int,
//         c: int,
//     }

s: Struct : {1, 2, 3}
// hover info:
//     s: struct {
//         a: int,
//         b: int,
//         c: int,
//     }
//
// should be "s: Struct : {1, 2, 3}"
```

And for procs:
```odin
foo :: proc (a, b: int) -> int
bar: proc (a, b: int) -> int
baz :: proc (a, b: int) -> int {return 0}

// currently all are displayed as "var: proc(a, b: int) -> int"

// where it should be
//     foo :: proc (a, b: int) -> int
//     bar: proc (a, b: int) -> int
//     baz: proc (a, b: int) -> int   // <- not sure if/how to include a value here

// maybe "baz :: proc (a, b: int) -> int ---" or "baz :: proc (a, b: int) -> int {...}"
```

thetarnav:
Not one of the usual "easy" issues unfortunately, but I thought there should be something on this.

BradLewis:
So just a note on the first point, `A` has the type `untyped integer` and `C` has the type `untyped float`. Right now they are just mapped to `int` and `float` respectively but we could easily change that to be `untyped integer` and `untyped float` which is the language the compiler uses.

As for the other points, I believe this would require some changes on how we internally resolve the symbols. Right now there isn't a distinction between something like `foo :: [2]f32` and `foo :: [2]f32{}`, which is something that is causing issues with the semantic tokens as well.

I'll spend some time on improving that and then we should be able to easily implement most of the changes here.

[...]

PR #1028, opened 2025-09-18
Feat/rework const hover info

BradLewis:
Will now show `Foo :: struct{}` rather than `Foo: struct{}` for types and `Foo : u8 : 16` rather than `Foo: u8` for constants.

Resolves https://github.com/DanielGavin/ols/issues/917

### DIFF
--- a/src/server/documentation.odin
+++ b/src/server/documentation.odin
@@ -249,6 +249,9 @@ write_signature :: proc(sb: ^strings.Builder, ast_context: ^AstContext, symbol: 
 		strings.write_string(sb, "}")
 		return
 	case SymbolProcedureValue:
+		if symbol.type == .Type_Function && depth == 0 {
+			strings.write_string(sb, "#type ")
+		}
 		write_procedure_symbol_signature(sb, v, detailed_signature = true)
 		return
 	case SymbolBitFieldValue:
@@ -342,6 +345,9 @@ write_short_signature :: proc(sb: ^strings.Builder, ast_context: ^AstContext, sy
 		write_node(sb, ast_context, v.value, "", short_signature = true)
 		return
 	case SymbolProcedureValue:
+		if symbol.type == .Type_Function {
+			strings.write_string(sb, "#type ")
+		}
 		write_procedure_symbol_signature(sb, v, detailed_signature = true)
 		return
 	case SymbolAggregateValue, SymbolProcedureGroupValue:
--- a/tests/hover_test.odin
+++ b/tests/hover_test.odin
@@ -182,7 +182,7 @@ ast_hover_procedure_with_default_comp_lit :: proc(t: ^testing.T) {
 		`,
 	}
 
-	test.expect_hover(t, &source, "test.fa :: proc(color_: Color = {255, 255, 255, 255})")
+	test.expect_hover(t, &source, "test.fa :: #type proc(color_: Color = {255, 255, 255, 255})")
 }
 
 @(test)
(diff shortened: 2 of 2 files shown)


thetarnav:
- proc types are displayed the same way as const procs
- constants that cannot be resolved completely are displayed as `CONST :: type` instead of `CONST : type : expression` or `CONST :: type(expression)` for example
- enum values are displayed a bit werid
  - if it has a explicit value: `package.Enum_Name: .Variant_Name = VALUE`
  - otherwise: `package.Enum_Name: .Variant_Name`
  - this is when hovering `.Variant_Name` btw
  - probably should be `package.Enum_Name.Variant_Name = VALUE` I guess
  - not sure if this is even relevant to the pr
- some constants are displayed differently between files:
  ```odin
  COLOR_BLUE :: frgba{0.1, 0.1, 1.0, 1.0}
  // hover in the same file is the same (with worse formatting)
  // hover in other files looks like this:
  //	main.COLOR_BLUE :: distinct [4]f32
  ```

I only given it 5 min of testing
but it works pretty great
and all but the last issues are not that important

BradLewis:
By worse formatting, you mean it's going over multiple lines rather than all on the same line?

But thanks that's really helpful!

thetarnav:
> By worse formatting, you mean it's going over multiple lines rather than all on the same line?

Yeah pretty much. But I guess that's expected. And not really an issue in any of my cases.

BradLewis:
I've pushed a fix for the issues across files so that should show correctly now. For the other points:

1. For the proc type vs proc declaration, I noticed this too but I'm honestly not sure how to distinguish these 2 cases in the hover info. Any ideas of how we could potentially do that?
2. I'm not entirely sure what you mean here, do you have a simple example?
3. That does sound like a general issue/improvement to be made with how the enum hover info is displayed rather than specific to this PR.

There are other issues which have been around for a bit where if you have something like this:

```odin
frgba :: distinct [4]f32

COLOR_BLUE := frgba{0.1, 0.1, 1.0, 1.0}

main :: proc() {
	foo := COLOR_BLUE
}
```
and you hover on `foo` it shows `foo: distinct [4]f32` rather than `foo: frgba` which would be more appropriate. Currently it will always resolve all the way down rather than just 1 level which I think is more appropriate, but this is again something I can look at later.

With this PR the completion label hasn't been updated, so it will still show `COLOR_BLUE: distinct [4]f32` as the label, but the docs will have the actual value.

thetarnav:
> 1. For the proc type vs proc declaration, I noticed this too but I'm honestly not sure how to distinguish these 2 cases in the hover info. Any ideas of how we could potentially do that?

[Odin docs do it by adding `{…}` at the end](https://pkg.odin-lang.org/core/slice/#reduce):

```
reduce :: proc(s: $/[]$U, initializer: $V, f: proc($V, $U) -> $V) -> $V {…}
```

It should work for any unknown or too long value.

> 2. I'm not entirely sure what you mean here, do you have a simple example?

It's the `C :: A + B` case.

Which has a additional bug now that I try it:

```odin
A :: 3
B :: 4
C :: A + B // C hover info: `main.C :: 4`
```

BradLewis:
> > 1. For the proc type vs proc declaration, I noticed this too but I'm honestly not sure how to distinguish these 2 cases in the hover info. Any ideas of how we could potentially do that?
> 
> [Odin docs do it by adding `{…}` at the end](https://pkg.odin-lang.org/core/slice/#reduce):
> 
> ```
> reduce :: proc(s: $/[]$U, initializer: $V, f: proc($V, $U) -> $V) -> $V {…}
> ```
> 
> It should work for any unknown or too long value.

That's similar to how we do it currently for completion labels with structs and whatnot. Should be easy to add here now as well.

> > 2. I'm not entirely sure what you mean here, do you have a simple example?
> 
> It's the `C :: A + B` case.
> 
> Which has a additional bug now that I try it:
> 
> ```odin
> A :: 3
> B :: 4
> C :: A + B // C hover info: `main.C :: 4`
> ```
Looks like there's still more work to do here then! 😅

BigBoyBarney:
Hi!

Would this show a constant's value as well?
For example currently:
image
image

BradLewis:
Yeah that's the plan. For anything constant it would show the value rather than trying to infer the underlying type.

BradLewis:
> * proc types are displayed the same way as const procs
> 
>     * constants that cannot be resolved completely are displayed as `CONST :: type` instead of `CONST : type : expression` or `CONST :: type(expression)` for example
> 
>     * enum values are displayed a bit werid
>       
>       * if it has a explicit value: `package.Enum_Name: .Variant_Name = VALUE`
>       * otherwise: `package.Enum_Name: .Variant_Name`
>       * this is when hovering `.Variant_Name` btw
>       * probably should be `package.Enum_Name.Variant_Name = VALUE` I guess
>       * not sure if this is even relevant to the pr
> 
>     * some constants are displayed differently between files:
>       ```odin
>       COLOR_BLUE :: frgba{0.1, 0.1, 1.0, 1.0}
>       // hover in the same file is the same (with worse formatting)
>       // hover in other files looks like this:
>       //	main.COLOR_BLUE :: distinct [4]f32
>       ```
> 
> 
> I only given it 5 min of testing but it works pretty great and all but the last issues are not that important

I've made corrections for these now (except for the enum stuff). For distinguishing between procs and proc types, I've opted to use the `#type` tag that odin has https://odin-lang.org/docs/overview/#type as I found that when using `{..}` it wasn't really standing out and was hard to miss as it's tucked all the way at the end of the definition (which for procs can be quite long). Using `#type` makes it very obvious that what you are looking at is indeed a type. Could always change it easy enough if desired.

PR #1040, opened 2025-09-22
Add new code action: remove unused imports

DanielGavin:
Add new code action to remove unused imports.


src/server/action.odin
```diff
@@ -67,11 +67,53 @@ get_code_actions :: proc(document: ^Document, range: common.Range, config: ^comm
...
 			add_missing_imports(&ast_context, selector, strings.clone(document.uri.uri), config, &actions)
 		}
+	} else if position_context.import_stmt != nil {
+		remove_missing_imports(document, strings.clone(document.uri.uri), config, &actions)
 	}
 
 	return actions[:], true
 }
 
+remove_missing_imports :: proc(
```
BradLewis, line 77: This should be `remove_unused_imports`?

BradLewis:
Works pretty well for me. Do you think it's possible to remove the lines entirely?

DanielGavin:
Good question! Was thinking that myself. I'll try and check.

src/server/imports.odin
```diff
@@ -1,17 +1,35 @@
...
+	_ = runtime.arena_init(&arena, mem.Megabyte * 40, runtime.default_allocator())
 
-	symbols_and_nodes := resolve_entire_file(document, .None)
+	defer runtime.arena_destroy(&arena)
 
+	context.allocator = runtime.arena_allocator(&arena)
 
+	symbols_and_nodes := resolve_entire_file_cached(document)
+
+	pkgs := make(map[string]bool, context.temp_allocator)
```
BradLewis, line 20: This is a minor thing but AFAIK you can do `map[string]struct{}` to create a map with a 0-sized value. (At least I'm pretty sure this is how it behaves in `odin`, it's like that in `go` at least 😅 )

src/server/action.odin
```diff
@@ -67,11 +67,65 @@ get_code_actions :: proc(document: ^Document, range: common.Range, config: ^comm
...
+				import_edit.range.start.character = column
+			}
+
+		}
+
+
+		append(&textEdits, import_edit)
+	}
+
+	log.error(textEdits[:])
```
BradLewis, line 111: Works for me! Just has this debug log here.

DanielGavin:
I needed to revamp the diagnostics system. You have to push everything all at once. We now have Unused, Syntax, Checker, diagnostics. 

It also was better for now to only check the unused on save and open.

BradLewis:
I'm getting a segfault when starting `ols` now with these most recent changes:

```sh
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x210)
    frame #0: 0x00000001026da82c ols`server::resolve_entire_file(document=0x0000000000000000, flag=None, allocator=runtime::Allocator @ 0x000000016d83cb70) at file_resolve.odin:83:2
   80           flag := ResolveReferenceFlag.None,
   81           allocator := context.allocator,
   82   ) -> map[uintptr]SymbolAndNode {
-> 83           ast_context := make_ast_context(
   84                   document.ast,
   85                   document.imports,
   86                   document.package_name,
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x210)
  * frame #0: 0x00000001026da82c ols`server::resolve_entire_file(document=0x0000000000000000, flag=None, allocator=runtime::Allocator @ 0x000000016d83cb70) at file_resolve.odin:83:2
    frame #1: 0x00000001026b1700 ols`server::find_unused_imports(document=0x0000000000000000, allocator=runtime::Allocator @ 0x000000016d83d140) at imports.odin:18:2
    frame #2: 0x00000001026d0188 ols`server::check_unused_imports(document=0x0000000000000000, config=0x00000001027f0388) at check.odin:75:2
    frame #3: 0x00000001026abe30 ols`server::notification_did_open(params=encoding_json::Value @ 0x000000016d83d740, id=<unavailable>, config=0x00000001027f0388, writer=0x000000016d83e1f0) at requests.odin:1046:2
    frame #4: 0x00000001026f84cc ols`server::call(value=encoding_json::Value @ 0x000000016d83db40, id=server::RequestId @ 0x000000016d83db28, writer=0x000000016d83e1f0, config=0x00000001027f0388) at requests.odin:347:4
    frame #5: 0x00000001026f3ae4 ols`server::consume_requests(config=0x00000001027f0388, writer=0x000000016d83e1f0) at requests.odin:291:3
    frame #6: 0x00000001026a8cd0 ols`main::run(reader=0x000000016d83e220, writer=0x000000016d83e1f0) at main.odin:83:3
    frame #7: 0x00000001026a90c4 ols`main::main at main.odin:128:2
    frame #8: 0x0000000102648c94 ols`main(argc=1, argv=0x000000016d83eab8) at entry_unix.odin:57:4
    frame #9: 0x0000000186962b98 dyld`start + 6076
```

DanielGavin:
Strange. 

It basically returns null on document for you.

```
if n := document_open(open_params.textDocument.uri, open_params.textDocument.text, config, writer); n != .None {
		return .InternalError
}

document := document_get(open_params.textDocument.uri)

check_unused_imports(document, config)
```

Doesn't happen on windows.

BradLewis:
Yeah it looks like it's failing to parse the `uri` in `document_get` and so it's returning `nil`

```odin
document_get :: proc(uri_string: string) -> ^Document {
	uri, parsed_ok := common.parse_uri(uri_string, context.temp_allocator)

	if !parsed_ok {
		return nil // returning here
	}

	document := &document_storage.documents[uri.path]
```

It doesn't seem which project I try I always get the same issue.

BradLewis:
Okay I see now, in `document_open`, the `uri_string` is deleted at the end of the function.

```odin
document_open :: proc(uri_string: string, text: string, config: ^common.Config, writer: ^Writer) -> common.Error {
	uri, parsed_ok := common.parse_uri(uri_string, context.allocator)

	if !parsed_ok {
		log.error("Failed to parse uri")
		return .ParseError
	}
...
	delete(uri_string)

	return .None
}
```

But in `notification_did_open` we have 

```odin
notification_did_open :: proc(
	params: json.Value,
	id: RequestId,
	config: ^common.Config,
	writer: ^Writer,
) -> common.Error {
	params_object, ok := params.(json.Object)
...

    // open_params.textDocument.uri deleted here
	if n := document_open(open_params.textDocument.uri, open_params.textDocument.text, config, writer); n != .None {
		return .InternalError
	}

     // open_params.textDocument.uri used again here!
	document := document_get(open_params.textDocument.uri)

	check_unused_imports(document, config)
...
}
```

DanielGavin:
Nice catch. Yeah probably not the smartest for it to be deleted in `document_open`

BradLewis:
LGTM! Everything works as expected now

PR #1048, opened 2025-09-23
Add proper support for ranged semantic tokens.

DanielGavin:


BradLewis:
I was playing around with this with vscode but I was encountering some weird behaviour: 

You can see the colours flashing

https://github.com/user-attachments/assets/051270df-eefe-44ab-8722-5159d6a18f32

PR #1049, opened 2025-09-23
Replace struct_field_value_by_name with custom find_derived

thetarnav:
I'm not exactly sure what this code does.
It looked like it searched the types for `ast.Expr.derived_expr` or `ast.Stmt.derived_stmt`.
Which could be rewritten a bit to replace comparing field names with field types.
It gives the same results based on my tests.
But I don't know if it doesn't miss something as everything still works even if I completely remove it.

### DIFF
--- a/src/server/clone.odin
+++ b/src/server/clone.odin
@@ -113,6 +113,7 @@ clone_node :: proc(node: ^ast.Node, allocator: mem.Allocator, unique_strings: ^m
 
 	res_ptr := reflect.deref(res_ptr_any)
 
+	// Recursively search for ast.Expr.derived_expr or ast.Stmt.derived_stmt field
 	find_derived :: proc(a: any) -> any {
 		if a == nil {return nil}
 


DanielGavin:
Seems like a better approach. Yeah it's hard to tell if there are issues.

BradLewis:
What's the advantage of this over using the builtin functionality? Performance improvements?

thetarnav:
yup, just performance.
I was curious if [this](https://discord.com/channels/568138951836172421/915622331383164968/1420123896027549716) could be easily improved.
I haven't profiled anything.
It just looked like an obvious change.
Although the reflection could probably be entirely eliminated if I took more time to understand it.

BradLewis:
I see. This code comes from the odin `clone_node` in `core:odin/ast` where they do the same thing: https://github.com/odin-lang/Odin/blob/master/core/odin/ast/clone.odin#L112

I've been testing this a bit on my end and haven't encountered any issues. Though as you noted that it doesn't seem necessary, I think we don't actually use `derived_stmt` or `derived_expr` at all in the code base, which would explain why. Given that we don't actually need it, maybe we can just remove it outright?

thetarnav:
It's being used in one place I think: https://github.com/DanielGavin/ols/blob/master/src/odin/printer/visit.odin#L257
Which I'm not sure if it could be written as`v.fullpaths[0].derived.(^ast.Basic_Lit)` as well

BradLewis:
Yeah that's just used for the formatter so it doesn't get cloned. Could also try using `derived` instead and see if it works as expected.

DanielGavin:
This code has been removed in https://github.com/DanielGavin/ols/pull/1089

PR #1053, opened 2025-09-24
Correct bool docs

BradLewis:


BradLewis:
These failing tests seem to be caused by the recent 2 commits that were added to odin, https://github.com/odin-lang/Odin/commit/5d3092bf2d00a46311a5f785658f04c823a9f3fa and the one before it.

PR #1060, opened 2025-09-26
Add inlay hints for implicit return values (closes #1058)

### LINKED ISSUE #1058: Add inlay hints for implicit return values
Adding inlay hints displayed after naked `return` could be a nice, unintrusive reminder that some existing variables are getting returned.

```odin
foo :: proc () -> (res: int, ok: bool) {

    if !condition() do return /* [[res, ok]] */

    condition() or_return /* [[res, <]] - not sure how to represent last return value */
    
    return value, true
}
```

There is an issue that "return value name" is a part of standard inlay hint types.
But I don't know if that matters for anything.

thetarnav:
- Adds inlay hints for implicit return values (closes #1058)
- Adds `enable_inlay_hints_implicit_return` config option for them.
- Removes `enable_inlay_hints` options, making it derived from the other inlay hint options.

Not sure if you like the change to the config.
It makes it more natural to me to just enable what you want to see.
Instead of enabling inlay hints in general, and then disabling what you don't want to see.


src/server/inlay_hints.odin
```diff
@@ -242,5 +267,67 @@ get_inlay_hints :: proc(
...
+		strings.write_string(&sb, " ")
+
+		for res, i in proc_data.results {
+			for name, j in res.names {
+				str := expr_name(name) or_continue
+				if i > 0 || j > 0 {
+					strings.write_string(&sb, ", ")
+				}
+				if is_or_return && i == len(proc_data.results)-1 && j == len(res.names)-1 {
+					strings.write_string(&sb, "⮡")
```
BradLewis, line 319: I wonder if it would be better to simply use `false` for this case?

thetarnav: I don't get it, you mean to just skip the last value from or_return?

thetarnav: I guess it could be `false` for booleans and name of the type for others

BradLewis: Ah yeah I didn't think of that. Maybe we could change it to be `_` instead. When testing on vscode, it simply looks like an empty square for me. I'm guessing I'm missing some utf-8 settings or something.
image

BradLewis:
Nice I think that looks pretty good!

PR #1062, opened 2025-09-27
Add a variable to set the odin branch/tag in the ci and nightly builds

BradLewis:
Given the recent instability of the master branch of odin, it'd be good to build off of the most recent stable release. This was just the simplest option I could think of for now. Should allow us to quickly update the version we use or even go back to master if things become more stable again.


DanielGavin:
Yeah seems like a fine idea.

PR #1063, opened 2025-09-27
Use the ranged argument from inlay hints to improve performance.

DanielGavin:
Noticed there was a parameter for range of inlay hints from the client. This should hopefully improve performance instead of getting inlay hints for the entire file.


BradLewis:
I seem to only get this error when trying this: 
image

PR #1075, opened 2025-10-01
Delete root package-lock.json

thetarnav:
I think this package-lock.json was created by mistake here, it has no contents. It's not necessary for anything, right?


DanielGavin:
Yeah that was a mistake. Thanks!

PR #1086, opened 2025-10-08
Avoid cloning AST nodes when collecting symbols

thetarnav:
Related to #1049
It avoids cloning the AST nodes when collecting symbols.
Previously the files were parsed using a temporary allocator, and then cloned to a collection allocator when collecting symbols.
Now the file is parsed using the collection allocator directly, and then just the strings in the AST are copied to the collection allocator later—the source code is freed after the request.

This works quite well and even makes tests run faster for me—from ~950ms to ~800ms.

### DIFF
--- a/src/server/clone.odin
+++ b/src/server/clone.odin
@@ -28,7 +28,11 @@ new_type :: proc($T: typeid, pos, end: tokenizer.Pos, allocator: mem.Allocator) 
 	return n
 }
 
-// Recursively deduplicate strings in AST nodes without cloning the structure
+/*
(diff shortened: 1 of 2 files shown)


BradLewis:
I had a play around with this and I noticed that the memory usage increases by a few megabytes everytime a save a file, even without changes. Saving the `analysis.odin` file increases the usage by ~7-8mb per save.

thetarnav:
right.. there should more memory used for the ast, and it's not freed entirely, as only ast nodes of symbols are freed, but that seems like a lot.

Issue #1059, opened 2025-09-26
Proc group inference in Odin got improved


thetarnav:
So this PR just got merged to Odin: https://github.com/odin-lang/Odin/pull/5689

It makes the following code compile, but ols does not support it yet:

```odin
package main

Enum :: enum {Foo, Bar, Baz}

group :: proc {proc_nil, proc_enum}
proc_nil  :: proc () -> int {return {}}
proc_enum :: proc (e: Enum) -> f32 {return {}}

main :: proc () {
    ret := group(.Foo)
/*   ^       ^     ^
     |       |     |
     |       |    No autocomplete for the Enum param
     |     `group` has no hover info
   `ret` has no hover info
*/ 
}
```

There might be more cases that this changes.

PR #1087, opened 2025-10-11
Feat/proc group implicit selector

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/1059, though it doesn't add support for the completions as I'm not sure how I would want to go about handling that.


DanielGavin:
Yeah the completion is tricky, since we can't confirm what function it is. We can just take all the functions and just give completion for all possible enums that are on the nth parameter the user is typing.

PR #1088, opened 2025-10-11
Update nightly release action to allow creating releases tied to odin versions

BradLewis:
It's been asked/discussed in the past to have 'stable' releases tied to odin versions that people can use. 

This PR updates/renames the nightly workflow to allow us to create releases based on the odin version. You create a tag such as `dev-2025-10` and it will create a new release with that name using that version of odin. 

The standard nightly process should work the same as normal. 

This releases won't really be 'stable' as `ols` in general is under active development, but it would mark points where people who are using older versions of odin would be able to use a version of `ols` that they can build.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,9 @@ jobs:
     - uses: actions/checkout@v1
     - name: Set ODIN_BRANCH from tag if available
       if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
-      run: echo "ODIN_BRANCH=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
+      run: |
+        echo "ODIN_BRANCH=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


DanielGavin:
Yeah seems good. I have also been wondering if we want an experimental branch that gets merged to master at some point. 

I really want people try https://github.com/DanielGavin/ols/pull/1049 as an example, but i'm also worried if it has problems.

BradLewis:
> Yeah seems good. I have also been wondering if we want an experimental branch that gets merged to master at some point.
> 
> I really want people try #1049 as an example, but i'm also worried if it has problems.

I noticed that we don't actually use those fields at all (`derived_expr` and `derived_stmt`) and so we don't actually need to clone them for now. I've created a PR for it here https://github.com/DanielGavin/ols/pull/1089.

But sounds good, I'll push a fix so that the version is updated in the binary correctly, and then I'll merge it in and create a release for `dev-2025-09`

BradLewis:
I've pushed a change to set the version in the binary. I don't have access to a windows machine at the moment to test the versioning, but it works fine on mac, and so should also on linux.

PR #1089, opened 2025-10-11
Remove cloning ast node `derived_expr` and `derived_stmt`

BradLewis:
After testing https://github.com/DanielGavin/ols/pull/1049, I noticed that we don't actually use these `derived_expr` nor `derived_stmt` at all and so we can remove the code outright and it should all work as expected. Will need to take note if we ever do require those fields so we can re-add it, but after testing for a while on this branch I haven't encountered any issues.

### DIFF
--- a/src/server/clone.odin
+++ b/src/server/clone.odin
@@ -2,15 +2,11 @@ package server
 
 import "base:intrinsics"
 
-import "core:fmt"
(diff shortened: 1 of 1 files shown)


DanielGavin:
Looks good. Yeah I doubt we would need to use those two fields.

PR #1105, opened 2025-10-17
os2 fixes getting executable path fail

KamWithK:
`get_executable_path` kept crashing for me on Linux (on latest git commit) so I fixed it by using the new function from `os2`


BradLewis:
Thanks! Looks good and works for me.

Issue #1125, opened 2025-10-28
ols does not update collections on second initialize request.


doongjohn:
`requests.odin -> proc read_ols_initialize_options`

```odin
// ...

	if "core" not_in config.collections && odin_core_env != "" {
		forward_path, _ := filepath.to_slash(odin_core_env, context.temp_allocator)
		config.collections[strings.clone("core")] = path.join(
			elems = {forward_path, "core"},
			allocator = context.allocator,
		)
	}

// ...
```

1. On the first initialize request: ols will insert base, code, vendor, shared in the `config.collections`.
2. On the second initialize request: ols will not insert the updated values.

This causes issues when a user specifies a custom `odin_command` in the `init_options`.
For example, in Neovim, the first initialize request is sent with the default parameters, followed by a second request containing the user’s custom `init_options`. Because of this, even if `odin_core_env` changes, the collections are not updated.

PR #1126, opened 2025-10-28
Fix read_ols_initialize_options order and collection overriding.

doongjohn:
Fixes: https://github.com/DanielGavin/ols/issues/1125

### DIFF
--- a/src/server/requests.odin
+++ b/src/server/requests.odin
@@ -542,8 +542,10 @@ read_ols_initialize_options :: proc(config: ^common.Config, ols_config: OlsConfi
 				}
 			}
 
-			if abs_core_env, ok := filepath.abs(odin_core_env, context.temp_allocator); ok {
-				odin_core_env = abs_core_env
+			if odin_core_env != "" {
+				if abs_core_env, ok := filepath.abs(odin_core_env, context.temp_allocator); ok {
+					odin_core_env = abs_core_env
+				}
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/server/requests.odin
```diff
@@ -539,34 +539,29 @@ read_ols_initialize_options :: proc(config: ^common.Config, ols_config: OlsConfi
 
 	log.infof("resolved odin root to: %q", odin_core_env)
 
-	if "core" not_in config.collections && odin_core_env != "" {
+	if odin_core_env != "" {
 		forward_path, _ := filepath.to_slash(odin_core_env, context.temp_allocator)
+
+		// base
+		config.collections[strings.clone("base")] = path.join(
```
BradLewis: We might need to free these cloned strings now that we're overriding the entries in the map.

BradLewis:
I tested this branch but I seem to have issues now where it can't find the collection specified within the `ols.json` file for the `ols` project:
image

This happens when I save the file.

doongjohn:
My assumption was wrong. Neovim does not send initialize request twice.

It was this code in `requests.odin`:
```odin
		read_ols_config(global_ols_config_path, config, uri)

		ols_config_path := path.join(elems = {uri.path, "ols.json"}, allocator = context.temp_allocator)

		read_ols_config(ols_config_path, config, uri)
		
		read_ols_initialize_options(config, initialize_params.initializationOptions, uri)
```
So it is calling `read_ols_initialize_options` 3 times. (read_ols_config calls read_ols_initialize_options)

So the last one passes the requested config `initialize_params.initializationOptions` which has no custom collections in it.

I think 3 calls is needed but in a different order:
1. load global config
2. load requested config
3. load ols.json

Edit:

`workspace/didChangeConfiguration` only give us the editor setting (without global and ols.json config) so we may need to do the 3 steps loading again when ols receives `workspace/didChangeConfiguration`.

doongjohn:
I didn’t make the default collections update when `odin_command` changes.
Should I make them update when `odin_command` changes?

BradLewis:
Seems to work well for me now, thanks!

Issue #1158, opened 2025-11-06
[Regression] `.` autocomplete shows all packages in a specific case


BigBoyBarney:
Hi!

I just discovered this regression in my project: [Mayling.odin, line 1061](https://codeberg.org/Sylphrena/Mayling/src/commit/54f2b913f2636a15d431cecc7c291f81ac99f8ff/Mayling/Mayling.odin#L1061).

With current `ols` master, using [Odin master](https://github.com/odin-lang/Odin/commit/40437b52ebca62b787337e1ddb684cbb92ff5b97) (as of the time of this writing).

<img width="1270" height="489" alt="Image" src="https://github.com/user-attachments/assets/24711f9f-51a5-4fb1-a509-0c5ad6621dca" />

The 2025-10 release does not have this issue:

<img width="1122" height="452" alt="Image" src="https://github.com/user-attachments/assets/9cf0c93e-5fea-48b4-8b1a-bc744e8eb2b5" />

----

Note: Extracting the line into a different file does not show the error. I could not reduce this to a small example, sorry :(

Cheers!

PR #1159, opened 2025-11-06
Correctly provide completions for an empty selector before a label

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/1158

As mentioned on discord, this was due to the selector being followed by a label. Since the label is added as part of the following statement, and the position is not in the statement itself, it was skipped.


BradLewis:
Closing as the larger change requires more odin parser improvements.

PR #1172, opened 2025-11-12
added #all_or_none

A1029384756:
this adds the `#all_or_none` directive. ~~this still doesn't show up in the errors pre-compilation but that seems to be because `odin check . -json-errors` reports `null` for the `pos` field so more work is probably going to be needed upstream~~ edit: fixed by pulling latest `master`


BradLewis:
Thanks! I didn't realise `all_or_none` had been added.

PR #1203, opened 2025-12-05
Add region folding for vscode

ctzcs:
Usually people perfer to write code in one file in odin. Region folding makes code more clearly. 
When someone reads the code, this lets them know what each piece is. And other languages generally have this feature.
```odin
//region name
//endregion
```

### DIFF
--- a/editors/vscode/language-configuration.json
+++ b/editors/vscode/language-configuration.json
@@ -28,5 +28,10 @@
         "increaseIndentPattern": "^.*\\{[^}\"']*$|^.*\\([^\\)\"']*$|^\\s*case.*:$",
         "decreaseIndentPattern": "^\\s*(\\s*\\/[*].*[*]\\/\\s*)*[})]"
     },
-    
+    "folding": {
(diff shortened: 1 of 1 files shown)


ctzcs:
If there is no one need this, i will close it.

xoxorwr:
I was looking for something like that, ping @DanielGavin

DanielGavin:
Sure seems like a fine idea.

reecewbourgeois:
+1 from me on wanting this feature. I initially was defining my own regions and then indenting everything within them:
```
// ** BEGIN Example Block ** //

    // Some code here
	
// ** END Example Block ** //
```

But then odinfmt removes the extra spacing within the "regions," so having it enabled in the language configuration would be nice.

BradLewis:
I can merge in the change though I'm unable to update the vscode extension. We'll need @DanielGavin  for that.

reecewbourgeois:
Much Appreciated! Just pulled it down and it works like a charm.

PR #1220, opened 2025-12-14
a few improvements and todo completions

TheBearodactyl:
Changes:
- Optimize comment retrieval (`ast.odin`)
- Provide context-specific directive completions (`completion.odin`)
- Improve upon `should_resolve_all_proc_overload_possibilities` (`analysis.odin`)
- Traverse nested fields when fuzzy searching (`memory_index.odin`)
- Remove position hint dependency from polymorphic type resolution (`generics.odin`)

### DIFF
--- a/src/server/ast.odin
+++ b/src/server/ast.odin
@@ -389,14 +389,12 @@ collect_value_decl :: proc(
 	stmt: ^ast.Node,
 	foreign_attrs: []^ast.Attribute,
 ) {
-	index := build_comment_index(file)
-	defer destroy_comment_index(&index)
 	value_decl, is_value_decl := stmt.derived.(^ast.Value_Decl)
 
 	if !is_value_decl {
 		return
 	}
-	comment, _ := get_file_comment(file, value_decl.pos.line, index = &index)
+	comment, _ := get_file_comment(file, value_decl.pos.line)
 
 	attributes := merge_attributes(value_decl.attributes[:], foreign_attrs)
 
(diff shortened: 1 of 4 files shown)


TheBearodactyl:
woah did not mean to do that 💀

TheBearodactyl:
one sec

TheBearodactyl:
fixed

TheBearodactyl:
image

welp, pretty sure the CI fails weren't me, so that's good

src/server/ast.odin
```diff
@@ -388,12 +389,14 @@ collect_value_decl :: proc(
 	stmt: ^ast.Node,
 	foreign_attrs: []^ast.Attribute,
 ) {
+	index := build_comment_index(file)
+	defer destroy_comment_index(&index)
```
BradLewis: I don't think this is "optimised" since it needs to build the index just to lookup one comment. It would be faster the way it is currently working.

BradLewis:
> image
> 
> welp, pretty sure the CI fails weren't me, so that's good

Actually since this working correctly on master I believe this is indeed from your changes. I have the same test failing on your branch when running locally.

TheBearodactyl:
oh, huh.
lemme try and figure out what's wrong
it is kinda weird that it's only failing on linux and macos though

src/server/ast.odin
TheBearodactyl: if you remove them it'll fallback to the original linear resolve

BradLewis:
It's actually failing on windows as well, but that doesn't fail the build 😅 
image

TheBearodactyl:
HUH

TheBearodactyl:
ok i have no idea what's going on because i haven't made any changes to the hover implementation and nothing i do fixes the test

Issue #1207, opened 2025-12-06
Implement source.organizeImports code action


Limero:
In Go (with gopls), I run the `source.organizeImports` code action on file save to automatically include all missing imports in the file, it would be useful to have the same in ols.

Behavior would be the same as the current `refactor.rewrite` action, but doing it for all missing imports, instead of just the one currently selected by the cursor.

Example for Neovim, this currently works for importing just the current package:
```
:lua vim.lsp.buf.code_action { context = { only = { 'refactor.rewrite' }, diagnostics = {} }, apply = true }
```
This is missing and would allow importing all at once:
```
:lua vim.lsp.buf.code_action { context = { only = { 'source.organizeImports' }, diagnostics = {} }, apply = true }
```

DanielGavin:
I'll look into it.

PR #1221, opened 2025-12-14
Add support for action: source.organizeImports

DanielGavin:
Adds requested feature from: [1207](https://github.com/DanielGavin/ols/issues/1207)


BradLewis:
How do you trigger this code action?

DanielGavin:
Vim:
`
:lua vim.lsp.buf.code_action { context = { only = { 'source.organizeImports' }, diagnostics = {} }, apply = true }
`
Vscode:
image

BradLewis:
Thanks. I don't seem to be able to get it to work in either neovim or vscode:
image

DanielGavin:
It only shows up if there are unused imports.

BradLewis:
Ah okay I see that works now. On the `go` side, `Organize Imports` will also add any missing imports.

DanielGavin:
Yeah seems like a good idea. It is a bit more work, but i'll try and see what I can do.

DanielGavin:
@BradLewis I have added support for removing and adding imports. It also supports top and bottom imports. In order to support for adding the imports, I needed to have a config for file resolve to also return unresolved symbols - currently only selector expressions.

BradLewis:
> @BradLewis I have added support for removing and adding imports. It also supports top and bottom imports. In order to support for adding the imports, I needed to have a config for file resolve to also return unresolved symbols - currently only selector expressions.

Nice that works pretty well for, however when I have a case where there's both a missing important and an unused import eg

```odin
package main

import "core:log"
import "core:fmt"


main :: proc() {
	fmt.println("hello!")
	fmt.println(strings.trim_space(" world "))
}
```

It looks like the addition and removal collide, I end up with

```odin
package main
ngs"
import "core:log"
import "core:fmt"

main :: proc() {
	fmt.println("hello!")
	fmt.println(strings.trim_space(" world "))
}
```

DanielGavin:
Thanks for testing!

This was something i also tested. This seems to work for me. I copied your code fully. 

I tried on vscode and nvim.

BradLewis:
Interesting, it also works for me on vscode but with neovim it consistently formats as mentioned above.

DanielGavin:
I have made the edits order independent. Some editors can go back to front or front to back.

BradLewis:
Perfect this works for me now! I think it's ready to go.

BradLewis (approved):

DanielGavin:
Tested on multiple editors and added tests.
  

Accidentally merged the branch. Used the GitHub mobile app. It showed an older comment where you said it worked - not sure why it doesn't jump to the latest.

Hopefully it works with you.

BradLewis:
All good, it's working fine for me.

PR #1228, opened 2025-12-17
Add defaultLibrary modifier for builtin

ske2004:
hello! This adds "defaultLibrary" modifier to `@builtin` declarations, to improve semantic highlight.

image


BradLewis:
Thanks!

PR #1230, opened 2025-12-21
Add config option to not align struct fields

lucypero:


DanielGavin:
Looks good.

You also need to update: https://github.com/DanielGavin/ols/blob/master/misc/odinfmt.schema.json

lucypero:
@DanielGavin done!

I also noticed that odinfmt aligns values when setting struct variables, too. Should I make a different PR/config parameter for that?

DanielGavin:
Yeah a separate configuration parameter makes sense.

You can add it in this PR.

lucypero:
Done! @DanielGavin

DanielGavin:
Great!

Issue #1165, opened 2025-11-07
Pack `odinfmt` on the tagged releases


flpStrri:
`odinfmt` is not exported to the tagged releases. Is there a reason for it not being there?

Sorry if this was already discussed, I could not find anything related on the repository.

flpStrri:
I made tons of copy pasta here to test it out: https://github.com/flpStrri/ols/commit/67a84a081eba6388b5694967c15367741a457de1

DanielGavin:
Seems like a fine idea. If you have already got it working, make a PR on it.

larzconwell:
@DanielGavin I'm about to make a similar PR that adds `odinfmt` to the release files, but I had a question about the inclusion of platform and architecture in the zipped binaries.

For context, here's my situation:
- I am using the following [Mise](https://mise.jdx.dev/) tool setup to install `ols`, which makes available on my machine the binary `ols-arm64-darwin`
  ```toml
  [tools."ubi:DanielGavin/ols"]
  provider = "github"
  version = "dev-2025-11"
  ```
- I am using [Helix](https://helix-editor.com/) which supports `ols` out of the box, but importantly it looks for a binary named `ols`

So unfortunately it's rather difficult to get a clean setup with `ols` currently as the binaries in the releases contain the platform and architecture meaning manual steps need to be taken to make it use the expected binary name.

Are you okay if in my PR that adds `odinfmt` to the releases, I also change it so the resulting binaries omit the platform and architecture? The zipped files will still contain the platform and architecture, so I don't think there's any loss of information.

DanielGavin:
I'm a bit worried about changing the binary names. I think there are quite a few editors already looking at the name. 

But I agree it might have been better if it was just named ols.

larzconwell:
Alrighty, will adjust my PR to forego the binary name changes and include `odinfmt` with the same naming structure used by the `ols` binaries so that they're at least consistent.

larzconwell:
@DanielGavin I have created a new PR that adds `odinfmt` without the change to the naming scheme.

https://github.com/DanielGavin/ols/pull/1231

flpStrri:
@larzconwell just in case you didn't saw it... you can make your Mise like this:

"ubi:flpStrri/ols[exe=odinfmt]" = "nightly" 
"ubi:DanielGavin/ols" = "dev-2025-10" 

The problem is that I failed to make it work with TWO bins using ubi

flpStrri:
[...]

PR #1231, opened 2025-12-21
Include odinfmt in release assets

larzconwell:
This builds `odinfmt` in CI along side `ols` and also updates the release workflow to include `odinfmt` in the zip files using the same filename scheme as `ols` for consistency.

Closes https://github.com/DanielGavin/ols/issues/1165


DanielGavin:
Thanks!

PR #1239, opened 2025-12-30
Avoid passing `-collection:shared=...` to odin invocation

FourteenBrush:
On ols version `dev-2025-12-30-16cd61e-debug` with odin version `dev-2025-12:c46e65809`, I am getting error logs because ols seems to pass a `-collection:shared=...` flag:

`Failed to unmarshal check results: Invalid_Data, Library collection 'shared' path must be a directory, got '/usr/local/bin/shared'`

The full odin invocation is:

```
odin check "/home/username/project/dir" -collection:shared="/usr/local/bin/shared" <checker_opts>  
 -json-errors 2>&1
```

I only seem to have this problem on the zed editor.

Also not entirely sure why this shared collection path is relative to the ols executable (basedir \<ols path\> / shared) (as per common.config.collections):

https://github.com/DanielGavin/ols/blob/16cd61e0f5ad5409f1747c24bdca545423892fdb/src/server/check.odin#L120

### DIFF
--- a/src/server/check.odin
+++ b/src/server/check.odin
@@ -118,7 +118,7 @@ check :: proc(paths: []string, uri: common.Uri, config: ^common.Config) {
 	collection_builder := strings.builder_make(context.temp_allocator)
 
 	for k, v in common.config.collections {
-		if k == "" || k == "core" || k == "vendor" || k == "base" {
+		if k == "" || k == "core" || k == "vendor" || k == "base" || k == "shared" {
 			continue
 		}
 		strings.write_string(&collection_builder, fmt.aprintf("-collection:%v=\"%v\" ", k, v))


FourteenBrush:
Actually seems to be a zed specific issue, as I did some further research, and it seems like zed passes a `RequestInitializeParams.initializationOptions` which contains collections, while nvim, vscode, among others pass an empty list for collections:

```hcl
	collections = [
		OlsConfigCollection{
			name = "core",
			path = "/usr/local/bin/core",
		},
		OlsConfigCollection{
			name = "shared",
			path = "/usr/local/bin/shared",
		},
		OlsConfigCollection{
			name = "vendor",
			path = "/usr/local/bin/vendor",
		},
```

This causes tabcompletion for the core, vendor and shared collections not to work.
I'll take a look at whats's wrong on the `zed-extensions/odin` side

FourteenBrush:
Zed issue has been fixed on version `v0.3.9` of the extension, seems like overriding the `shared` collection is actually permitted, which can not be said about core, vendor and base, I'll close this now then.

https://github.com/zed-extensions/odin/commit/3176fc17a2e1a5ad6f62ac26d64371ef7a05ef77

PR #1240, opened 2025-12-30
added stdout flag to odinfmt

xandaron:
Added `stdout` flag to odinfmt to allow outputting to stdout. Is only valid when also using stdin. I considered just making this the default behaviour when using stdin instead of fmt.println but decided to leave that decision to someone else.


xandaron:
It was pointed out to me that fmt.println uses stdout so this is pointless.

Issue #481, opened 2024-08-17
package proc autocomplete bug in sublime text


thetarnav:
```odin
import "core:slice"

slice.clon // ...
```

in vscode accepting `clone` autocompletion produces `slice.clone`, but in sublime it produces just `clone`, replacing the `slice.` part. It only happens with procedues, not with variables or types.

It happens regardless of `enable_procedure_snippet` setting. (the parenthesizes are not added in both cases, even in local procedures)

haven't debug it enough to tell if this is a sublime/ols bug tbh
but it happens only in odin

DanielGavin:
I can repro it. But looking at the completion lsp it doesn't actually tell sublime to replace it. There is no difference between completion that works and doesn't from the lsp log. It most likely has to be sent to sublime LSP.

PR #1244, opened 2026-01-04
fix `in`/`not_in` bit_set complete

ske2004:
Re-add missing dot in this bit set completion:

```odin
Pokemon :: enum {
  Missingno,
  Snorlax,
  Jigglypuff,
}

Pets :: bit_set[Pokemon]

main :: proc() {
  my_pets := Pets{}
  
  if my_pets.S // <- here 
}
```

Before:
```
  if Snorlax in my_pets // <- here 
```

After:
```
  if .Snorlax in my_pets // <- here 
```

### DIFF
--- a/src/server/completion.odin
+++ b/src/server/completion.odin
@@ -835,6 +835,9 @@ get_selector_completion :: proc(
 		remove_edit, rok := create_remove_edit(position_context, true)
 		if !rok {break}
 
+		// Sublime Text will remove the original `.` for some reason
+		is_sublime := config.client_name == "Sublime Text LSP"
+
 		for name in enumv.names {
 			append(
 				results,
@@ -855,7 +858,7 @@ get_selector_completion :: proc(
 						label = fmt.tprintf(".%s in", name),
 						kind = .EnumMember,
 						detail = in_text,
-						insertText = in_text,
+						insertText = is_sublime ? in_text : in_text[1:],
 						additionalTextEdits = remove_edit,
 					},
 				},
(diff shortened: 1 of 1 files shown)


BradLewis:
What editor and OS is this for? This change causes it to insert an extra `.` on macos with zed and vscode. Eg 
image

ske2004:
@BradLewis Hm, this is Sublime's doing, that's pretty strange.

ske2004:
I added a quick fix for this, but I understand if this hackery isn't acceptable.

BradLewis:
Yeah I can reproduce this in sublime as well. I'm guessing there something weird going on with the code that determines what gets removed.

ske2004:
Yeah there's all sorts of weirdness with Sublime LSP https://github.com/DanielGavin/ols/issues/481 :/

BradLewis:
Yeah that wouldn't be the first sublime specific code we have for the completions. I had a play around with it to see if I could find something that worked for all editors, but I wasn't able to. I think we just merge this in as it does what we need. Thanks!

ske2004:
Thanks! Sublime has some strange behaviours around `.`, if anything like this happens again I will try to find out why it does that instead of submitting hacks. I hadn't had similar issues in clangd so I wonder what it does differently. I also tried to play around with different insert modes but it seems like sublime simply strips the leading dot (and in some cases, as we've seen with #481, leading identifier before it). It's possible Sublime or Sublime-LSP is making some assumptions about the language syntax.

Issue #1245, opened 2026-01-05
[Bug]: Build fails with error "_ of type struct has no field 'is_all_or_none'


sebastian-j-ibanez:
### Issue

Build fails on [latest commit ](af7ece508418a62cd85ca6c453738d81cda68306)of master branch with 3 errors:
```sh
$ ./build.sh
/home/sebas/Code/ols/src/server/collector.odin(169:5) Error: 'struct_type' of type '^Struct_Type' has no field 'is_all_or_none'
        if struct_type.is_all_or_none {
           ^~~~~~~~~~^

/home/sebas/Code/ols/src/server/symbol.odin(433:6) Error: 'v' of type '^Struct_Type' has no field 'is_all_or_none'
        if v.is_all_or_none {
           ^

/home/sebas/Code/ols/src/odin/printer/visit.odin(1616:6) Error: 'v' of type '^Struct_Type' has no field 'is_all_or_none'
        if v.is_all_or_none {
```

### Solution

The build was successful after deleting these three if statements that referenced the field `is_all_or_none`. 

### Odin version
```
odin version dev-2025-11:e5153a937
```

BradLewis:
This is because you are using an older version of odin that doesn't have the all_or_none changes, which was added in dev-2025-12. You'll need to update your odin or if you want to stick with dev-2025-11 you can use the dev-2025-11 tag of ols.

PR #1246, opened 2026-01-05
Remove broken references to 'is_all_or_none' field

sebastian-j-ibanez:
### Changes made
Remove bad references to `is_all_or_none` field.

Fixes https://github.com/DanielGavin/ols/issues/1245.


BradLewis:
As mentioned in the issue, this is valid odin and required for the formatter. You just need to update your odin.

sebastian-j-ibanez:
The Arch package must be out of date. I will look into it

Issue #1248, opened 2026-01-08
odinfmt truncates large array literals


mollstam:
I have the byte array of the font inlined in Odin code (yeah buddy) and now when I ran odinfmt on the project it stopped compiling:

```Error: Expected 80000 values for this array literal, got 10919```

Turns out odinfmt tossed the first 68k bytes of the data, here's an example file to run odinfmt on that will get corrupted, haven't looked into it much further but though I wanted to create an issue first.

Here's a Python script to generated the Odin source that will get trouble from odinfmt, sorry for not making it a Odin script, I'm not that cool.

```
block_size = 400
blocks = 200
size = block_size * blocks

with open('will_truncate_when_formatted.odin', 'w') as f:
    f.write("package main\n")
    f.write("\n")
    f.write(f"my_bytes: [{size}]u8 = {{\n")

    for b in range(blocks):
        for _ in range(block_size):
            f.write(f"    {b},\n")

    f.write("}\n")
    f.write("\n")
    f.write("main :: proc() {}\n")
```

mollstam:
All right, been digging around a bit wrapping my head around this AST formatting and printing code to find were things goes wrong. Seems to in the end be just that the arena allocator used gets full and starts returning Out of Memory errors which aren't handled and thus the formatting will just go with what it has.

In my case its visiting a Document_Cons node and the list it appends to is at `len` and `cap` `131064`. It tries to resize the allocation with the arena to `8388352` bytes which fails. The arena has 50 MB to work with so not sure if its fragmentation or legit usage from having such a large document (this append being the straw breaking the camel's back). Changing the arena to have 100 MB and I can format all fine.

What's worrisome is that I first setup a test for this large array literal which _does not_ fail, but running `odinfmt` manually on the test case fails, so there might be future work to make the tests suite's allocator more closely match what's used from CLI to support catching these kinds of things.

DanielGavin:
I think it's best just to switch over to a virtual arena allocator - the same one that is used by the temporary allocator.

PR #1251, opened 2026-01-12
Use default heap allocator instead of fixed 50 MB arena for odinfmt

mollstam:
Root cause of my issue #1248 (barring having a Very Large Source File) was that the 50 MB arena allocator wasn't enough. I don't think its much of a solution to just up the number and I don't want to get into the headache of dynamic arrays and arenas, so the simple solution is just not use a custom arena and instead take from the heap as much as needed...

Sadly peak memory used statistic can no longer be tracked but maybe that is OK.

Sorry for a bit of a "cold drop" and let me know if you have any thoughts. :pray:

### DIFF
--- a/tools/odinfmt/main.odin
+++ b/tools/odinfmt/main.odin
@@ -46,7 +46,7 @@ walk_files :: proc(info: os.File_Info, in_err: os.Errno, user_data: rawptr) -> (
 
 main :: proc() {
 	arena: vmem.Arena
-	arena_err := vmem.arena_init_growing(&arena, 50)
+	arena_err := vmem.arena_init_growing(&arena)
 	ensure(arena_err == nil)
 	arena_allocator := vmem.arena_allocator(&arena)
 


DanielGavin:
Think it's better to switch to a growing allocator.

```
import vmem "core:mem/virtual"

// This creates a growing virtual memory arena. It uses virtual memory and
// can grow as things are added to it.
arena: vmem.Arena
arena_err := vmem.arena_init_growing(&arena)
ensure(arena_err == nil)
arena_alloc := vmem.arena_allocator(&arena)


```

mollstam:
Hi Daniel, thanks for taking a look. I've only used Odin a little bit over a long period of time so my knowledge isn't that deep but in this case of using a vmem arena shouldn't we make it the static variant, doesn't the growing one cause pointer invalidation if it can't grow in place when it fills up its block (1 MB default)? Seeing as the already allocated memory would get moved to a new larger block. So maybe make it the static one with a big reservation should do the trick?

DanielGavin:
The growing allocator keeps the blocks in a linked list, so when the allocator grows it will not invalidate any data previously allocated.

mollstam:
All right, seems to work as desired with the vmem arena. ✨

PR #1252, opened 2026-01-14
Feat: Add support for odinfmt custom config path

gxhamster:
Introduces a new argument to -config which allows a user to pass a json config for the formatter which is not neccesarilly in the path of the files we are formatting. For now the function will use the default_style if any errors occur. This was done due to that being the pattern in the source code. I think it would be useful to know if reading a config has failed or has defaulted to a default style but, I wanted to not include that in this. Maybe in the future this can be addressed in another pull request. There has been a few occasions where I noticed that the formatter was not using the config that I was expecting.

### DIFF
--- a/src/odin/format/format.odin
+++ b/src/odin/format/format.odin
@@ -49,6 +49,26 @@ find_config_file_or_default :: proc(path: string) -> printer.Config {
 	return config
 }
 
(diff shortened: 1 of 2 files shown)


DanielGavin:
Thanks!

PR #1254, opened 2026-01-17
Improve hover documentation formatting

louisdutton:
## Summary

- Merged `get_doc` and `get_comment` for standardised comment aggregation and presentation
- Delimit hoverdoc sections with horizontal rules, similar to many other LSP implementations
- Removed odin codeblock wrapping the whole hoverdoc (yes some internal docs will no longer be highlighted but it's arguably better to meet convention and allow arbitrary markdown codeblocks in doc comments.)
- Updated expected content in hover tests

## Example

image

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -1981,8 +1981,8 @@ resolve_local_identifier :: proc(ast_context: ^AstContext, node: ast.Ident, loca
 	return_symbol.flags |= {.Local}
 	return_symbol.value_expr = local.value_expr
 	return_symbol.type_expr = local.type_expr
-	return_symbol.doc = get_doc(local.docs, ast_context.allocator)
-	return_symbol.comment = get_comment(local.comment)
+	return_symbol.doc = get_comment(local.docs, ast_context.allocator)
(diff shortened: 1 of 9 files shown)


BradLewis:
Thanks for this! I think it looks really nice. I'm just doing some testing on my end to make sure that everything still works as expected. I noticed an issue with the documentation with raylib. Before when hovering over a function it would look like this: 
image
Now it seems we have lost the comment:
image

louisdutton:
Thanks for taking a look @BradLewis . I don't seem to be able to reproduce this issue. which editor are you using?
image

BradLewis:
This is neovim. I think this is actually an existing issue rather than one you introduced though. It's definitely finicky but if you start with a simple example like this:

```odin
package main

import rl "vendor:raylib"

main :: proc() {
	rl.WindowShouldClose()
}
```
There's no comments:
image

This includes with completions:
image

However if you have the raylib source file open when it first parses the file, the comments will indeed appear just fine as you showed here: 
image

But as mentioned I can reproduce this now on master so I'll fix that on my end. But looks good for your PR so thanks again!

PR #1268, opened 2026-01-29
Invert if Code Action

pnarimani:
Added a code action to invert an if statement without changing the logic

This will help reducing nesting in the code.

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -1,3 +1,4 @@
+#+feature using-stmt
 package odin_printer
 
 import "core:fmt"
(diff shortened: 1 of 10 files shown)


BradLewis:
Seems to work well for me. One thing I noticed is that in vscode it'll always have this icon now for any `if` statement you have in your code:

image

I know in the past people have commented on that being annoying. Is it possible to only have the action show when highlighting or something like that?

DanielGavin:
I think it would have to be fixed on the client.  You can disable the light bulb in vscode: 
```
 "editor.lightbulb.enabled": "off"
```
Don't think there is a way to tell it not to show for a specific source action.

pnarimani:
I checked a C# project with VS Code and that also shows a lightbulb even when inverting the if statement is not suggested.

BradLewis:
Alright, well it seems to work well for me! If people start mentioning the vscode thing we can take a look at it then 😅 . But thanks again! Good stuff!

PR #1273, opened 2026-01-31
feat: code action to populate remaining switch cases

DireLines:
https://github.com/user-attachments/assets/7b3306b3-9891-4f2b-92c9-06735bcba1d7

notes:
- does not keep indentation as you can see in demo. feel free to block on that
- detecting and keeping the code for cases which are misnamed does not work unless it parses as an implicit selector. feel free to block on that

### DIFF
--- a/src/server/action_populate_switch_cases.odin
+++ b/src/server/action_populate_switch_cases.odin
@@ -27,27 +27,32 @@ get_block_original_text :: proc(block: []^ast.Stmt, document_text: string) -> st
 }
 
 SwitchCaseInfo :: struct {
-	name:             string,
+	names:            []string,
 	body_indentation: string,
 	body:             string,
+}
+SwitchBlockInfo :: struct {
+	existing_cases:         []SwitchCaseInfo,
+	all_covered_case_names: []string,
+	all_case_names:         []string,
+	switch_indentation:     string,
+	is_enum:                bool,
 }
 get_switch_cases_info :: proc(
 	document: ^Document,
 	ast_context: ^AstContext,
 	position_context: ^DocumentPositionContext,
 ) -> (
-	existing_cases: []SwitchCaseInfo,
-	all_case_names: []string,
-	switch_indentation: string,
-	is_enum: bool,
+	info: SwitchBlockInfo,
 	ok: bool,
 ) {
 	if (position_context.switch_stmt == nil && position_context.switch_type_stmt == nil) ||
 	   (position_context.switch_stmt != nil && position_context.switch_stmt.cond == nil) {
-		return nil, nil, "", false, false
+		return {}, false
 	}
 	switch_block: ^ast.Block_Stmt
 	found_switch_block: bool
+	is_enum: bool
 	if position_context.switch_stmt != nil {
 		switch_block, found_switch_block = position_context.switch_stmt.body.derived.(^ast.Block_Stmt)
 		is_enum = true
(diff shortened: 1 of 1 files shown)


BradLewis:
Thanks for this! There's some conflicts now due to the other action PR merged in, but I noticed that it reorders the cases. Do you think it would be possible to make it so it only adds the missing cases at the end of the list of existing cases? Would also make the indentation less jarring too.

DireLines:
yep, will do

DireLines:
done! indentation is still jarring, do you know a way to keep each line's original indentation in the text replacement?

BradLewis:
Thanks! [@pnarimani](https://github.com/pnarimani) recently added a proc that calculates that in their PR here: https://github.com/DanielGavin/ols/pull/1268/changes#diff-460fe6493518a2043a27aca2e460d016178a6fcea6e874e5abd61dfdccf0e15eR241

You might be able to use that for an existing `case` statement or the `switch` statement itself?

DireLines:
ok, handled indentation @BradLewis

BradLewis:
Awesome that works great! I just noticed that for type switch statements it misses any leading package declaration or points from the type. So if it's supposed to be `^ast.Value_Decl`, it'll instead just be `Value_Decl`.

DireLines:
ok, I will see how the autocomplete suggestions detect that and copy the logic

DireLines:
done. I also noticed that the logic breaks if you do multiple cases in one case block(`case .a, .b:` and the like). Will see if there's a comprehensive solution for that, or at least detect whether all the cases are covered if the programmer used those and if so just don't offer the code action

DireLines:
ok, you can do range syntax for enums, so this throws a wrench in things. the compiler at least knows which cases aren't covered, so if I can use the results of the odin parser I don't need to duplicate that logic

```ts
main :: proc() {
	E :: enum {
		a,
		b,
		c,
		d,
		e,
		f,
		g,
	}
	e: E
	switch e {
	case E(0) ..= E(3):
	}
}
```

DireLines:
darn, also appears not to handle parapoly union variants properly. Will return to this in a while and make test cases and fix this stuff, but I'm just gonna leave it for now

BradLewis:
I'll take a look on my end to see if I can get the `case .A, .B:` working. I think after that it'll be pretty good to go! Especially since its opt-in and the other cases don't seem very likely.

DireLines:
@BradLewis handled multiple comma-separated variants.

briefly taking a look at the parapoly union case that broke when I attempted using the action on somebody's library code. Might give up on fixing that if it's an annoying enough issue

DireLines:
here's the code that breaks if you wanna take a look (stripped down version of the library code, so it doesn't mean anything)

```ts
package main
Save_Game :: struct($Game: typeid) {
	game_rollback_state: ^Game,
}
Load_Game :: struct($Game: typeid) {
	game_rollback_state: ^Game,
}
Skip_Frames :: struct {
	num_frames: int,
}
Advance_Frame :: struct($Input: typeid) {
	inputs: [5]Input,
}
Rollback_Request :: union($Game, $Input: typeid) {
	Save_Game(Game),
	Load_Game(Game),
	Advance_Frame(Input),
	Skip_Frames,
}
main :: proc() {
	Game :: distinct int
	Input :: distinct int
	request: Rollback_Request(Game, Input) = Skip_Frames{}
	//this is what we want - should detect there are no more cases to cover here
	switch &r in request {
	case Save_Game(Game):
	case Load_Game(Game):
	case Advance_Frame(Input):
	case Skip_Frames:
	}
}
```

BradLewis:
I've got it working mostly working now, but it seems to not work correctly in the example you provided for hover as well.
image

I'll take a look at that but seems unrelated to this code action.

I also did a bit of refactoring so now it'll only add the missing cases rather than rewriting the whole switch statement. 

I think I'll merge in this PR now, and then merge in my changes separately, followed by looking into the issue you found. Thanks again!

PR #1276, opened 2026-02-05
Fix alias resolution for package imports

xoxorwr:
Fixes a bug where package aliases were not properly resolved, preventing completion suggestions for members of aliased packages

Check the test for example

**Disclaimer: This was vibe coded**


BradLewis:
It seems to work okay for me. Could you clean up the comments and maybe add a test for the `foo :: foo` case that was mentioned in the original issue?

xoxorwr:
Oops that was meant to stay on the fork

I'll send proper cleaned PR once ready

PR #1281, opened 2026-02-09
fix: forgot dot

DireLines:


DireLines:
@BradLewis noticed a bug introduced by the refactor of `get_qualified_union_case_name`

BradLewis:
Ah whoops thanks!

PR #1315, opened 2026-02-16
fix vs code "Debug Test" button

mvikto:
Vs code "debug test" button doesn't do anything on my laptop. I noticed that `odin debug test . -debug` runs that test executable and immediately deletes it. Adding -keep-executable seems to fix the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ cd ols
 ./odinfmt.sh
 ```
 
-In order for `ols` to find symbols for builtin types and procedures, the `builtin` folder in the repo needs to be located next to the `ols` binary.
+In order for `ols` to find symbols for builtin types and procedures, the `builtin` folder in the repo needs to be located next to the `ols` binary. Alternatively you can specify the path to this folder using the `OLS_BUILTIN_FOLDER` environment variable.
 
 ### Configuration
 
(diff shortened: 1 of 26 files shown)


BradLewis:
Does this directory get cleared by vscode or something like that? I can imagine it'll pile up with executables if it doesn't.

mvikto:
when I do `odin test --help` I see this flag:
```
        -keep-executable
                Keep the executable generated by `odin test` or `odin run` after running it. We clean it up by default.
                If you build your program or test using `odin build`, the compiler does not automatically execute
                the resulting program, and this option is not applicable.
                
```

It looks like odin removes that executable. I assume that cleanup was a recent change to odin cli? Otherwise, I can't see how this debug button worked before at all?
I don't think pile up is a concern.  `odtest odin test . --debug -keep-executable` produces/overwrites executable which has the same name as the directory where it's run. So you only have executable from the recent test run.

mvikto:
hmm it implies I can `odin build` a test instead of relying on `odin test -debug` to produce executable. We should perhaps do that instead?

BradLewis:
I seem to be getting this error on mac when trying to debug some tests using the changes mentioned here: 
image

Is this the error that you were seeing?

mvikto:
I was testing on hello world project not on ols.  Then initially I got 0 feedback until opened dev console and saw that "Not possible to find executable, candidates are:" log.

 After playing around with ols tests I made some more changes until it worked on my laptop.

I also had to add  collections thing in local ols.json
```json
{
    // ....
    "collections": [
        { "name": "src", "path": "./src"}
    ]
  }
```

BradLewis:
I just did a test on my end with your latest changes and it seems to be working now! 
image
Is it all working for you as well?

mvikto:
It seems be working for me in ols repo, yes. But I noticed some problem with code lense regex not quite working correctly on a sample project. I think I will try to fix address in another pr though.

BradLewis:
I can merge this in then but we'll need @DanielGavin to create a release for the extension.

PR #1317, opened 2026-02-17
Update intrinsics.odin

xoxorwr:
This file was outdated

I didn't bother running the formatter, makes updating easier, plain copy pasta


BradLewis:
The parser doesn't like the where clauses since there's no procedure body. I'm not sure if that would cause issues?

xoxorwr:
Let's fix the parser then, i'll take a look

BradLewis:
I did some testing on my end and even with all the parser errors it seems to work fine for me, and even better now since it includes a lot more information in the description. It'll also be a lot easier to maintain since it's just copy-paste, so thanks!

PR #1333, opened 2026-02-25
Load global ols.json even when no project uri is provided

Znarf64:
It seems that helix sometimes not send neither a "rootUri" nor any "workspaceFolders" (specifically when there is no root file it can find, by default main.odin or ols.json). This causes the global default ols.json to not be respected. I'm not sure to what extent this should be considered helix's fault, but I don't think there's a reason not to load the global config in that case.


BradLewis:
Thanks!

PR #1341, opened 2026-02-27
Fix stray trailing comma in ols.schema.json

Sprial404:
Hi team,

I noticed that the commit (c8842e3), introduced three (3) days ago, added a new section "checker_skip_packages" which added a trailing comma. I have introduced a pull request to amend it since I noticed that the schema was no longer being parsed without error in my text editor. 

Thank you for your time in advance.


BradLewis (approved):
Thanks!

PR #1351, opened 2026-03-07
Add support for new for init; x in y {} syntax

RcCreeperTech:
# Add support for `for init; x in y {} syntax`
Odin dev-2026-03 introduced an optional initializer statement in for-in loops:
```odin
for x := make_iterator(); v in x {
    // ...
}
```
This patch updates OLS to correctly handle the new init field on Range_Stmt across five files:

- `src/server/locals.odin` walk stmt.init so variables declared in the initializer are visible inside the loop body
- `src/server/position_context.odin` include stmt.init when resolving document position, so completions and hover work inside loops that have an init
- `src/server/file_resolve.odin` resolve stmt.init so symbols declared there are correctly indexed
- `src/server/ast.odin` free stmt.init to avoid a memory leak
- `src/odin/printer/visit.odin` emit the init statement and trailing ; when formatting a Range_Stmt that has an init

Tests covering the new behavior are added in tests/completions_test.odin.

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -1197,6 +1197,10 @@ visit_stmt :: proc(
 			document = cons(document, visit_expr(p, v.label), text(":"), break_with_space())
 		}
 
+		if v.init != nil {
+        	document = cons(document, visit_stmt(p, v.init), text(";"), break_with_space())
+    	}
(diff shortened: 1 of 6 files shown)


BradLewis:
Thanks!

Issue #1366, opened 2026-03-16
[BUG] Segfault on code using unions


pawelt:
Valid Odin code causes segfault. First line works, second line doesn't:

```
    // elsewhere: SysError :: os.Error

    if x == 2 do return SysError(mem.Allocator_Error(.Invalid_Pointer))
    if x == 2 do return SysError(.Invalid_Pointer)
```

Output from ols:
```
...
026-03-16 17:11:40.205 [error] Starting Odin Language Server nightly-2026-03-15-9e0b036
2026-03-16 17:11:40.468 [error] Server process exited with signal SIGSEGV.
2026-03-16 17:11:40.469 [info] Connection to server got closed. Server will restart.
2026-03-16 17:11:40.472 [error] Starting Odin Language Server nightly-2026-03-15-9e0b036
2026-03-16 17:11:40.715 [error] Server process exited with signal SIGSEGV.
2026-03-16 17:11:40.716 [error] The Odin Language Server Client server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.
```

BradLewis:
Thanks for raising this! Let me know if you encounter other issues.

PR #1367, opened 2026-03-16
Fix segfault with poly unions

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/1366


BradLewis:
Seems to be an issue on the odin master branch? All passes fine on my side.

PR #1377, opened 2026-03-21
Private struct fields that start with underscore

### LINKED ISSUE #414: [FEAT] private struct variables
Would it be possible to make it so that the language server wouldnt suggest certain struct variables/procedures outside the current package or file? Like marking the variable with an underscore or @(private) like we already have for procedures and global variables. 

Wouldnt be 100% needed, just a little feature. And yeah maybe i should ask ab this on the official language github, but dont need them to actually be private so you cant access them, just so that you wouldnt 'see' them.

Marco012:
Added a new config option called "enable_private_struct_fields_underscore" that hides struct fields that have an underscore prefix out of the package where the struct is declared.

Closes #414

Had to remove ast.Fixed_Capacity_Dynamic_Array_Type from clone.odin to compile it, probably because of different Odin version.

### DIFF
--- a/src/server/completion.odin
+++ b/src/server/completion.odin
@@ -604,8 +604,6 @@ get_comp_lit_completion :: proc(
 	config: ^common.Config,
 ) -> bool {
 
-	current_package := ast_context.current_package
-
 	if symbol, ok := resolve_comp_literal(ast_context, position_context); ok {
 		#partial switch v in symbol.value {
 		case SymbolStructValue:


BradLewis:
Yeah you'll need to keep that `Fixed_Capacity_Dynamic_Array_Type` code otherwise it'll crash when encountering those ast nodes 😅. You'll just need to update your odin version for it to compile.

Marco012:
Ah I apologize, I thought I was the one with the most recent version.
I was also looking forward to implement warnings for when we try to access the struct fields outside the package, although I'm having a hard time figuring out where that would be. I tried implementing it on resolve_selector_expression proc on the analysis.odin file. Is this wrong?

BradLewis:
I'm not sure `ols` actually reports any errors itself, it's mostly just forwarding the errors from the odin parser or when running `odin check` so I'm not sure where would be appropriate for something like that.

Marco012:
Ah ok, that explains it. I'll leave it as that then. Thanks.

BradLewis:
Thanks!

PR #1401, opened 2026-03-28
Update README.md code snippets to use jsonc

eliavaux:
There are two JSON code snippets that use comments which are not allowed in standard JSON. Thus the syntax highlighter marks these sections in red. JSONC functions exactly the same way but allows comments and is probably better suited for these two snippets.

It's a little pedantic I admit, but they do look nicer now.


BradLewis:
Thanks!

PR #1415, opened 2026-04-06
Implement `convert_do` for `odinfmt`

ej-uses-git:
Hi! Thank you for all your hard work on OLS and all the associated tools. As far as I understand, the `convert_do` option currently doesn't work - this PR implements it within the code.

I've confirmed that this works with running the compiled binary, but I notice there's no tests for `odinfmt`. Would you like me to create them and add a test for this behavior? Or should I leave a TO-DO comment somewhere to inform future maintainers that we'd like to test this?

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -944,7 +944,7 @@ visit_stmt :: proc(
 	case ^ast.Using_Stmt:
 		document = cons(document, cons_with_nopl(text("using"), visit_exprs(p, v.list, {.Add_Comma})))
 	case ^ast.Block_Stmt:
-		uses_do := v.uses_do
+		uses_do := v.uses_do && !p.config.convert_do
 		is_single_line := v.open.line == v.end.line
 
 		if v.label != nil {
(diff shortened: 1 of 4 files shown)


BradLewis:
Thanks! There actually are tests in `tools/odinfmt/tests` for `odinfmt`. You can look at `tools/odinfmt/tests/single_line_switch/` to see how you can set it up to run with some configuration (you can see the `odinfmt.json` in there). It works by formatting the file and comparing to the expected output in `.snapshots`

ej-uses-git:
> Thanks! There actually are tests in `tools/odinfmt/tests` for `odinfmt`. You can look at `tools/odinfmt/tests/single_line_switch/` to see how you can set it up to run with some configuration (you can see the `odinfmt.json` in there). It works by formatting the file and comparing to the expected output in `.snapshots`

Huh - no idea how I missed that! Will write some tests and add them to this PR.

ej-uses-git:
Okay, I've added a very simple test that ensures that it works. Would you like me to add others?

BradLewis:
Thanks!

PR #1427, opened 2026-04-16
Fix missing trailing comma in proc types

### LINKED ISSUE #1425: [Bug] odinfmt breaks trailing comma in long proc type returns
When a long proc type with a single return value is formatted by odinfmt, there is no trailing comma after the multi-lined return value (and if there was one, it is removed). This only happens for procedure types and not for normal proc declarations (which is inconsistent but not really a problem in itself). 

The issue is that `odin check` with `-strict-style` errors if there is no trailing comma in multi-line statements, including after a single return value.
```odin
// -strict-style wants a trailing comma after the return type
dangerousType :: #type proc(
        my_very_wordy_and_very_long_argument: int,
) -> (
        my_extremely_important_and_sadly_very_wordy_result: int
)

// with multiple return types, both get a comma and everything is fine
fineType :: #type proc(
        my_very_wordy_and_very_long_argument: int,
) -> (
        my_extremely_important_and_sadly_very_wordy_result: int,
        another_result: bool,
)

Example :: struct {
        // typed fields also don't get a comma after the return (and -strict-style still wants one)
        field: proc(
                my_very_wordy_and_very_long_argument: int,
        ) -> (
                my_extremely_important_and_sadly_very_wordy_result: int
        ),
}

// for proc definitions, everything works as expected
normalProc :: proc(
        my_very_wordy_and_very_long_argument: int,
) -> (
        my_extremely_important_and_sadly_very_wordy_result: int,
) {
         return 0
}
```

PhoenixPhantom:
Fixes #1425

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -2236,7 +2236,7 @@ visit_proc_type :: proc(
 			document = cons_with_nopl(document, text("("))
 			document = cons(
 				document,
-				nest(cons(break_with(""), visit_signature_list(p, proc_type.results, contains_body, true))),
+				nest(cons(break_with(""), visit_signature_list(p, proc_type.results, true, true))),
 			)
 			document = cons(document, break_with(""), text(")"))
 		} else {


BradLewis:
Thanks!

PR #1431, opened 2026-04-18
adds #simple

Nave55:
I was trying to use #simple and noticed it wasn't supported by ols. Looked at a previous pr for directives and tried to replicate it.


BradLewis:
Thanks! It looks like I missed that that was added recently.

BradLewis (approved):

PR #1432, opened 2026-04-19
Fix multilined proc type return values

### LINKED ISSUE #1428: [Bug] Broken formatting for multiline proc return value
While testing my changes in #1427, I noticed that something like the following snippet is formatted incorrectly. I was first thinking this was caused by my changes, but it seems that the behavior was already present before the PR, so I'm opening a separate issue here.

```odin
// the comma after the return type on the same line as the `{` is a problem
my_medium_long_proc_name :: proc(
	$x: int,
) -> proc(
		my_very_wordy_and_very_long_argument: int,
		my_other_very_wordy_and_very_long_argument: int,
	) -> bool, {
	return {}
}
```

One solution would be to have the formatting add brackets like the following or simply to omit the comma 
```odin
my_medium_long_proc_name :: proc(
	$x: int,
) -> (
       proc(
		my_very_wordy_and_very_long_argument: int,
		my_other_very_wordy_and_very_long_argument: int,
	) -> bool, 
) {
	return {}
}
```
I'll try looking into a fix during the weekend, but I can't guarantee I'll find the time for it.

PhoenixPhantom:
Fixes #1428

I've decided to go with the extra brackets variant I suggested in the issue primarily because I prefer how it looks, but I also found it easier to implement.

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -2211,6 +2211,7 @@ visit_proc_type :: proc(
 		document = cons(document, text(">"))
 
 		use_parens := false
+		can_multiline_single := false
 
 		if len(proc_type.results.list) > 1 {
 			use_parens = true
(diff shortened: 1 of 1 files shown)


DanielGavin:
Seems good. Tested on my project without without any formatting issues. Thanks!

PR #1433, opened 2026-04-22
make `checker` thread_local

Faker-09:
Should the global variable  `checker` in check.odin be thread_local?


BradLewis:
I don't think so no. `queue_check_request` should be called by multiple threads to queue requests. Right now that's not the case but that's the idea.

PR #1440, opened 2026-05-01
build: fix build error on recent Odin nightly (Essence OS target)

4bdullah0:
This PR fixes a build error encountered when using recent Odin nightly 
versions (e.g., dev-2026-04)

The compiler now requires #partial or an explicit case for the new 
'Essence' OS target in enumerated arrays. Added #partial to 
os_enum_to_string in src/server/build.odin to restore compatibility.

Error fixed:
Unhandled enumerated array case: Essence

### DIFF
--- a/src/server/build.odin
+++ b/src/server/build.odin
@@ -34,7 +34,7 @@ platform_os: map[string]struct{} = {
 }
 
 
-os_enum_to_string: [runtime.Odin_OS_Type]string = {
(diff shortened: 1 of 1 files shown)


BradLewis:
Actually this was a target that was recently removed, not added, in the compiler. See https://github.com/DanielGavin/ols/pull/1439

Updating your odin version should fix this issue.

PR #1449, opened 2026-05-07
Resolve symbols to their base type when comparing for overload function checks

### LINKED ISSUE #1447: make() with allocator argument (other than context.allocator) breaks rename/references
## Description

OLS fails to find references (and thus rename) for local variables declared with `make([dynamic]T, allocator)` when the allocator argument is anything other than `context.allocator`. This affects variable rename, reference-finding, and likely other cross-reference features.

The underlying cause appears to be that OLS can't resolve the type of `make()`'s return value when a non-`context.allocator` allocator is passed, which breaks reference tracking for the assigned variable.

## Steps to reproduce

1. Create the following Odin file:

```odin
package repro

import "core:fmt"
import "core:mem"

main :: proc() {
	// OK: context.allocator directly — references found
	{
		x := make([dynamic]int, context.allocator)
		fmt.println(len(x))
	}

	// BROKEN: nil mem.Allocator local — 0 references
	{
		a: mem.Allocator
		y := make([dynamic]int, a)
		fmt.println(len(y))
	}

	// BROKEN: default_allocator() — 0 references
	{
		z := make([dynamic]int, mem.default_allocator())
		fmt.println(len(z))
	}
}
```

2. Place cursor on `y` (or `z`) in the declaration `y := make(...)` and invoke rename or find-references.
3. Observe that OLS returns 0 references / no rename edits for `y` and `z`, but works correctly for `x`.

## Expected behavior

All three variables should have references found (2 references each: declaration + usage in `fmt.println(len(...))`).

## Actual behavior

- `x` (`context.allocator`): 2 refs found ✅
- `y` (`a: mem.Allocator` nil local): 0 refs found ❌
- `z` (`mem.default_allocator()`): 0 refs found ❌

## Additional context

This also affects `mem.arena_allocator()` return values passed to `make()`, which is a common pattern in real-world Odin codebases (e.g. frame allocators in game loops).

The bug is reproducible even when `textDocument/prepareRename` returns a valid range — OLS identifies the symbol at the cursor but fails to resolve its references.

## Environment

- OLS version: nightly-2026-05-03-992787a7 (installed via Mason nvim)
- Odin version: dev-2026-04-nightly:a896fb2
- OS: Linux

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/1447

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -374,6 +374,33 @@ are_symbol_basic_same_keywords :: proc(a, b: Symbol) -> bool {
 	return true
 }
 
+// Fully resolve a symbol to it's base.
+// Used for comparing that two symbols are the same when checking overloads
+resolve_base_symbol :: proc(ast_context: ^AstContext, symbol: Symbol) -> Symbol {
+	if .Distinct in symbol.flags {
+		return symbol
+	}
+
+	expr := symbol.type_expr
+	if expr == nil {
+		expr = symbol.value_expr
+	}
+	if expr == nil {
+		return symbol
+	}
+
+	set_ast_package_from_node_scoped(ast_context, expr^)
+
+	resolved, ok := resolve_type_expression(ast_context, expr)
+	if !ok {
+		return symbol
+	}
+	if resolved.name == symbol.name && resolved.pkg == symbol.pkg && resolved.type == symbol.type {
+		return symbol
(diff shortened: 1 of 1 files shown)


BradLewis:
Hmm this causes issues elsewhere. Back to the drawing board I think.

Issue #1422, opened 2026-04-15
[Bug] Alias in another file break overload resolution of functions in another package


DarathDev:
main.odin
``` odin
package ols_error
import "my_package"

Bar :: proc {
	my_package.Bar,
	my_package.Bar2,
}

// a's and b's hover tooltips do not work
main :: proc() {
	a := Bar(Foo(int){})
	b := Bar(Foo2{}, int)
}
```

other.odin
``` odin
package ols_error
import "my_package"

Foo :: my_package.Foo
Foo2 :: my_package.Foo2
```

my_package/package.odin
``` odin
package my_package

Foo :: struct($T: typeid) {}
Foo2 :: struct {}
Bar :: proc(table: Foo($T)) -> (data: T) {return}
Bar2 :: proc(table: Foo2, $T: typeid) -> (data: T) {return}
```

DarathDev:
Here is a simpler minimal reproduction

main.odin
``` odin
package hover_error
import "my_package"

Bar :: proc {
	my_package.Bar,
}

// Bar's tooltip does not work
main :: proc() {
	Bar(Foo{})
}

```

other.odin
``` odin
package hover_error
import "my_package"

Foo :: my_package.Foo
```

my_package/package.odin
``` odin
package my_package

Foo :: struct {}
Bar :: proc(table: Foo) {}
```

DarathDev:
@BradLewis I tried debugging this myself by adding a test case, but I am not sure if I can represent a second file in the default package by making an entry in the package array with the default name `test`. From what I can tell it is `Foo{}` that cannot be resolved be resolved, but that might just be a mistake in my test setup.

BradLewis:
I've added some very basic support for testing across multiple files in this MR https://github.com/DanielGavin/ols/pull/1490. That should hopefully allow you to properly set up a test case. If you're unable to get it to work I'll try and have a look at it this weekend.

DarathDev:
Turned out that I could replicate it even use multiple packages as well. I think I have a suitable fix for it.

PR #1467, opened 2026-05-15
Test ignore file tags

thetarnav:
- Adds support for multiple files in tests for both main package and imported packages.
- Improves ignore file tags tests by testing both between file and between packages behaviors.

In the meantime while this was a draft @BradLewis already added a support for multiple files in #1490 (which was very confusing when rebasing), so this just offers a different interface (maybe slightly more intuitive) to the same thing.

### DIFF
--- a/src/testing/testing.odin
+++ b/src/testing/testing.odin
@@ -7,6 +7,7 @@ import "core:odin/ast"
 import "core:odin/parser"
 import "core:strings"
 import "core:testing"
+import "core:path/filepath"
 
 import "src:common"
 import "src:server"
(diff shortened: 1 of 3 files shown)


BradLewis:
> In the meantime while this was a draft @BradLewis already added a support for multiple files in https://github.com/DanielGavin/ols/pull/1490 (which was very confusing when rebasing), so this just offers a different interface (maybe slightly more intuitive) to the same thing.

Ah yeah I just did that very quickly to help with https://github.com/DanielGavin/ols/issues/1422, though it wasn't even used in the end. More than happy to change it to a better API.

BradLewis (approved):

PR #1493, opened 2026-05-22
Fix issue resolving arguments to an overloaded proc that are aliases

DarathDev:


DarathDev:
Resolves #1422

src/server/analysis.odin
```diff
@@ -402,14 +402,18 @@ resolve_base_symbol :: proc(ast_context: ^AstContext, symbol: Symbol, bypass_dis
...
 		return symbol
 	}
-	if resolved.name == symbol.name && resolved.pkg == symbol.pkg && resolved.type == symbol.type {
-		return symbol
+	if  resolved.type == symbol.type {
+		if resolved.name == symbol.name && resolved.pkg == symbol.pkg{
+			return symbol
+		}
+		// If the type stayed the same we shouldn't resolve all the way to an anonymous struct
+		resolve_base_symbol(ast_context, resolved, bypass_distinct, false)
```
BradLewis: I actually think this block is all you need. You can remove all the `to_anonymous` stuff. 

Edit: actually that's weird that even works. I just realised you're not returning this, but yeah having 

```odin
		resolve_base_symbol(ast_context, resolved, bypass_distinct)
		return resolve_base_symbol(ast_context, resolved, bypass_distinct)
```
makes it pass for me...

BradLewis: It seems all we need is:

```odin
	if  resolved.type == symbol.type {
		if resolved.name == symbol.name && resolved.pkg == symbol.pkg{
			return symbol
		}
		// If the type stayed the same we shouldn't resolve all the way to an anonymous struct
		reset_ast_context(ast_context)
	}
```

That works for me as well.

DarathDev: > Edit: actually that's weird that even works. I just realised you're not returning this, but yeah having

Ah, I forgot to return, but it worked so I didn't notice. 

> It seems all we need is: ...

This works for this case, but then it assumes all same base type values are equivalent when choosing candidates for the overload. 

Edit: Instead of preventing the call arg from degenerating to an anonymous struct it makes the expected arg also degenerate, which seems wrong.

The below example fails to compile anyway, so if its fine for incorrect code to have the wrong tooltips, then it probably works.

### Counter Example

pkg1/pkg1.odin
```odin
package pkg1

Foo :: struct {}
Baz :: struct {
	a: int,
}
Bar :: proc(table: Foo) {}
```

pkg2/pkg2.odin
```odin
package pkg2
import "../pkg1"

Foo :: pkg1.Foo
Baz :: pkg1.Baz
```

main.odin
```
package test
import "pkg1"
import "pkg2"

Bar :: proc {
	pkg1.Bar,
}

// Bar's tooltip is pkg1.Bar :: proc(table: Foo)
main :: proc() {
	Bar(pkg1.Baz{})
}
```

BradLewis: Hmm yeah I see the same behaviour when adding `return resolve_base_symbol(ast_context, resolved, bypass_distinct, false)` as well. When removing the `to_anonymous` and keep the resolve without the return 

```odin
	if resolved.type == symbol.type {
		if resolved.name == symbol.name && resolved.pkg == symbol.pkg{
			return symbol
		}
		resolve_base_symbol(ast_context, resolved, bypass_distinct)
	}
```
it works for me.

In general though we're okay with resolving incorrect types for programs that don't compile. As far as I can tell as well, the `.Anonymous` flag is only used when displaying the documentation so I don't think it's needed here anyways and makes sense that it doesn't make a difference with the resolution.

DarathDev: I will use `reset_ast_context(ast_context)` as its the change with the smallest footprint and resolves the current issue. If you are fine with it I think it is ready to be merged.

I still suspect it might cause a regression elsewhere so I added an inline note. Maybe `AstContext` should have a field for resolving to anonymous values?

BradLewis: This is only used as the last attempt to resolve a proc overload so it should be okay. If you do find any other issues though we can come back to it. Thanks!

PR #1496, opened 2026-05-23
fix issue resolving  overload with polymorphic arguments when passed …

### LINKED ISSUE #1495: [Bug] Poly overloads fail to resolve when passed a return which is a pointer
```odin
package test

Foo :: struct($T: typeid) {}

bar :: proc(foo: Foo($T)) -> ^T {
	return {}
}

baz :: proc {
	baz_pass,
}

baz_pass :: proc(value: $T) -> T {
	return T{}
}
// baz's tooltip doesn't work
main :: proc() {
	foo: Foo(struct {})
	baz(bar(foo))
}

```

DarathDev:
…a pointer to a generic from another function

Attempts to fix #1495 

The problem seems to be that during `resolve_generic_function_symbol` the parameter resolves to the type without the pointer, as `symbol_to_expr` doesn't retain that information. I am not certain if that is intended. The changes in this pull request change that, but it breaks several other tests, and causes the server to crash. 

Is there another mechanism by which this information should normally be retained, or is it an oversight of `symbol_to_expr`?

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -251,7 +251,7 @@ visit_decl :: proc(p: ^Printer, decl: ^ast.Decl, called_in_stmt := false) -> ^Do
 			}
 			document = cons(document, text("}"))
 		} else if len(v.fullpaths) == 1 {
-			if _, ok := v.fullpaths[0].derived_expr.(^ast.Basic_Lit); ok {
+			if _, ok := v.fullpaths[0].derived.(^ast.Basic_Lit); ok {
 				document = cons_with_nopl(document, visit_expr(p, v.fullpaths[0]))
 			} else {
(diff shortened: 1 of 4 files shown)


BradLewis:
I'm not 100% sure, but if it's breaking other tests like that it's probably intended to work like this. My guess is it's something like it's not correctly recursing down to realise it should be a `^Bar` and not just a `Bar`.

This should be a similar test case that demonstrates the same problem which should make debugging easier.

```odin
package test

Bar :: struct{}

bar_ptr :: proc(value: $T) -> ^T {}

bar_pass :: proc(value: $T) -> T {}

main :: proc() {
	b := bar_ptr(Bar{})
	c := bar_pass(b)
}
```

BradLewis:
Thanks! Works well for me.

PR #1499, opened 2026-05-28
Add inlay hints for unhandled optional results

thetarnav:
Another useful inlay hint idea.
It adds a hint about missed optional result.
Right now it only works for proc calls with `#optional_ok` or `#optional_allocator_error`, so no map access or union casts.
Also only works on declarations and assignments as idk how to handle inline calls.


BradLewis (approved):
Thanks!

PR #1500, opened 2026-05-31
fix(build): Error in build pointing to missing exhaustive case in os_enum_to_string for Haiku.

MauriceElliott:


MauriceElliott:
Closed as this was rolling back the future and completely unnecessary, more an issue with my own Odin version.

PR #1506, opened 2026-06-02
scoped const semantic token fix

### LINKED ISSUE #1494: Wrong semantic tokens for constants defined in scope without explicit type
```odin
TRUE :: bool(true) // @lsp.type.variable.odin

some_proc :: proc () {
    TRUE :: bool(true) // @lsp.type.type.odin
}
```
Happens both for `Type(val)` and `Type{val}`

thetarnav:
Fixes #1494 and the `max_u32 :: max(u32)` hover case by unifying a bit how local and global resolving works.


BradLewis:
Thanks!

PR #1512, opened 2026-06-09
Support fake methods with proc aliases

### LINKED ISSUE #1018: Proc aliases are not included in fake methods
```odin
Thing :: struct {a, b: int}

thing_foo :: proc (t: ^Thing) {}
thing_bar :: thing_foo

thing: Thing

thing. /* autocomplete: (with fake methods enabled)
 - a
 - b
 - thing_foo
    < should also include thing_bar
*/
```

Similar: #485

### LINKED ISSUE #1486: Fake Methods not working correctly with procedure aliases
Fake methods are not being detected when they are aliased by a separated module, here is a small example:

- `core/math/math.odin`: Considering a simple rect struct with a function to return its area
```odin
package math

Rect :: struct {
    width: f32,
    height: f32,
}

rect_area :: proc(r: Rect) -> f32 {
    return r.width * r.height
}
```

- `core/core.odin`: Then a parent module, defines an alias for the `rect_area` function
```odin
package core

import "math"

Rect :: math.Rect
rect_area :: math.rect_area
```

This way the rect_area method doesn't show up when using in `main.odin`:

Image

But if instead of aliasing the `rect_area` function i do this in `core/core.odin`:
```odin
rect_area :: proc(r: Rect) -> f32 {
    return math.rect_area(r)
}
```

It shows up correctly:

Image

However, if I create another alias (or even a function wrapper), the auto completion will use the function from the `core` package instead of the local one.

Image

thetarnav:
closes #1018
closes #1486

\+ made the `expect_completion_labels` error output nicer to read.

### DIFF
--- a/src/server/collector.odin
+++ b/src/server/collector.odin
@@ -504,6 +504,45 @@ get_or_create_package :: proc(collection: ^SymbolCollection, pkg_name: string) -
 }
 
 /*
+	Collects a fake method from an alias if references a proc.
+*/
+collect_alias_method :: proc(collection: ^SymbolCollection, symbol: Symbol, visited: ^map[^Symbol]struct {}) {
+
+	pkg := &collection.packages[symbol.pkg]
+	value := symbol.value.(SymbolGenericValue)
+
+	resolved: ^Symbol
+	#partial switch v in value.expr.derived {
+	case ^ast.Ident:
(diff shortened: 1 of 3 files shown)


BradLewis:
Thanks! That's a much nicer output for the errors. Going to make it so much easier to scan what's missing.

PR #1513, opened 2026-06-11
fix: add missing Haiku OS case to fix compilation with latest dev compiler

Marineseeker:
### Description
The latest nightlies/dev versions of the Odin compiler introduced the `.Haiku` member to `runtime.Odin_OS_Type`. This caused `os_enum_to_string` (which is an enumerated array) to fail compilation due to an unhandled case.

This PR adds the missing `.Haiku` enum case and its corresponding string mappings to `src/server/build.odin`, allowing OLS to be successfully compiled using the latest Odin compiler.

### DIFF
--- a/src/server/build.odin
+++ b/src/server/build.odin
@@ -28,9 +28,8 @@ platform_os: map[string]struct{} = {
 	"wasi"    = {},
 	"wasm"    = {},
 	"netbsd"  = {},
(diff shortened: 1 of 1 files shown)


BradLewis:
Haiku was removed recently, not added. You'll need to either update your Odin version or use a pinned release of ols.

PR #1518, opened 2026-06-12
Include workspace root in package list for workspace symbol search

### LINKED ISSUE #1517: Workspace symbols miss odin files in workspace root directory
workspace/symbol returns no symbols from .odin files placed directly in the workspace root. Moving the same files into a subdirectory makes them appear.  I've looked into the code and it seems that only subdirectories of the workspaces are searched.

TxFig:
Fixes #1517.

Odin files in the workspace root directory were not included in workspace/symbol results because only subdirectories were collected when building the package list. The root path is now added explicitly.


BradLewis (approved):
Thanks!

PR #1525, opened 2026-06-14
Clarify `enable_fake_methods` and `enable_procedure_snippet` require …

bennyyip:
Clarify `enable_fake_methods` and `enable_procedure_snippet` require client snippet support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Options:
 
 - `enable_document_symbols`: Turns on outline of all your global declarations in your document. _(Enabled by default)_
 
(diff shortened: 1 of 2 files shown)


BradLewis:
Was this causing you issues with a particular editor?

bennyyip:
I am using https://github.com/yegappan/lsp, which has no built-in snippet support. I tried `fake_method` and it doesn't work. I was confused, so it is better to clarify that these features need snippet support.

BradLewis:
Makes sense, thanks!

PR #1539, opened 2026-06-29
fix segfault updating configuration with no workspaces

xdBronch:
been running into this with nvim, i had `root_markers` set to `'.git'` but there was no git repo in my project so the workspace was empty


BradLewis (approved):
Thanks!

PR #1541, opened 2026-07-01
Adds compress_values and expand_values

Nave55:
Adds compress_values and expand_values to builtin and tests.


BradLewis (approved):
Thanks!

PR #1548, opened 2026-07-08
Remove \r on `.sh` files, and prevent git from readding them

rats159:
Currently all of the shell scripts are a little broken, running them produces errors like `‘bash\r’: No such file or directory`, because the shell doesn't consider \r to be whitespace I guess? This PR removes the \r's, and adds a .gitattributes file to prevent git from readding them, specifically for shell files

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sh text eol=lf


BradLewis:
Was there supposed to be changes in the actual shell files?

rats159:
Yeah, I think git wound up removing them again. Maybe I needed the attributes to be in a commit beforehand? If you pull down the PR does it remove them, or is git just ignoring the attributes file?

EDIT: I think the changes went through properly, I see them when I pull it down. Maybe GitHub doesn't show line ending changes?

BradLewis:
On my machine there are no `\r`'s in any of these. As far as I can tell this is the same for the committed files. Do you have some git settings that automatically converts this into `\r\n` ? Maybe something like `git config core.autocrlf` ?

rats159:
Yeah, that is enabled for me. Apparently it's the default on windows, so the issue may have come from the fact that I cloned ols on windows but then used it through wsl. I believe that's a fairly common workflow though, so it may still be worth keeping the attribute file around in case other people also clone on windows but use wsl?

BradLewis:
Yeah that seems like a good idea. Should stop other people having this issue or even accidentally committing this. Thanks!

PR #1555, opened 2026-07-12
Find Odin files faster in Windows

lucypero:
Makes odin file search much faster in windows. This search is done when renaming symbols or finding references.

My testing on https://github.com/lucypero/lucydx12:

Before PR: `295ms`
After PR: `42ms`

Revert f0c5301feed55823e541155349fe1b8ec28c3a92 to log timings.

### DIFF
--- a/src/common/util.odin
+++ b/src/common/util.odin
@@ -8,6 +8,7 @@ import "core:os"
 import "core:path/filepath"
 import "core:strings"
 import "core:time"
+import "core:slice"
 
 foreign import libc "system:c"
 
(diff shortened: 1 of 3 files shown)


BradLewis:
Nice! This has made a pretty substantial difference on my windows machine, going from ~15ms to ~4ms for `ols`, ~140ms to ~24ms on the `odin` repo and ~150ms to ~30ms on your `lucydx12` repo. Another thing I noticed while testing is with the old version I would sometimes get big spikes on the file lookup (usually ~1s but sometimes >5s and one time it was over 20s) which never happened with your changes.

PR #1557, opened 2026-07-14
remove an extra trailing newline caused by `fmt.println` in odinfmt

FyLost:
I noticed that neovim and helix always make sure there are two `\n`s at end of file on saving, because `odinfmt -stdin` is used in a typical configuration. If using lsp as the formatter (`vim.lsp.buf.format()` in neovim), then there would be only one newline at end of file. The extra newline is caused by `fmt.println` in odinfmt. (I am not sure if it is intended for odinfmt)

### DIFF
--- a/tools/odinfmt/main.odin
+++ b/tools/odinfmt/main.odin
@@ -90,7 +90,7 @@ main :: proc() {
 		)
 
 		if ok {
-			fmt.println(source)
+			fmt.print(source)
 		}
 
 		write_failure = !ok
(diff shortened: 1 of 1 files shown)


BradLewis:
What os is this for? I have the same behaviour with or without this change when running `vim.lsp.buf.format()` on macos 😅 . I'm just trying to reproduce this on my end.

FyLost:
Because this `fmt.println` is only called by the cli tool `odinfmt`, not by `vim.lsp.buf.format()`. If you configure formatting on save following the [README](https://github.com/DanielGavin/ols#neovim), you will see that two newlines are put at end of file because of the `fmt.println`. (I am on Linux)

BradLewis:
Ah I see, thanks!

PR #1559, opened 2026-07-15
Correct semantic token for generic procs

### LINKED ISSUE #1558: Couple of semantic token issues
Few noticed semantic token issues to correct

```odin
generic_proc :: proc ($FOO: int) {
    //                  ⤷ `FOO` has no token

    FOO // `FOO` has a `@lsp.type.typeParameter.odin`
        // even though it's not a type
        // probably it's hardcoded for `$T: typeid` case
}
```

```odin
Foo :: struct {
	callback: proc ()
}
foo: Foo
foo.callback()
//     ⤷ `callback` is @lsp.type.property.odin
//        while it should probably be @lsp.type.function.odin
```

BradLewis:
Resolves https://github.com/DanielGavin/ols/issues/1558


thetarnav:
This misses the readonly flag so const params get the same highlight as other params.

PR #1566, opened 2026-07-17
adds type inference support for compress and expand values

Nave55:
I've added type inference support for compress and expand values, including the "**" operator for expand values. 
I did quite a bit of testing to make sure there are no apparent bugs. One issue that I wasn't able to solve was how constants work in compress values. For compress values to work correctly it needs to be able to determine if a value is typed or untyped. Constants should be seen as untyped but I couldn't figure out an easy way of receiving this info. So in certain circumstances with literals of the same type in compress values but differing on whether or not they are untyped, the hover will be wrong.

### DIFF
--- a/src/server/builtins.odin
+++ b/src/server/builtins.odin
@@ -424,8 +424,9 @@ get_expr_return_value :: proc(
 		return get_expr_return_value(ast_context, i.expr, true, allocator)
 
 	case ^ast.Basic_Lit:
-		i.tok.text = token_kind_to_str(i.tok)
-		return i, fmt.tprintf("u %v", i.tok.text), 0, true
+		ident := new_type(ast.Ident, n.pos, n.end, allocator)
+		ident.name = token_kind_to_str(i.tok)
+		return ident, fmt.tprintf("u %v", ident.name), 0, true
 
 	case:
 		if v, v_ok := resolve_type_expression(ast_context, n); v_ok {
@@ -449,6 +450,7 @@ get_symbol_return_value :: proc(
 	return_len: int,
 	ok: bool,
 ) {
+	untyped := false
 	ast_context.use_locals = true
 
 	#partial switch sym_val in symbol.value {
@@ -458,17 +460,21 @@ get_symbol_return_value :: proc(
 
 	case SymbolUntypedValue:
 		if sym_val.type == .Bool {
+			untyped = .Mutable in symbol.flags == false ? false : true
 			ident := new_type(ast.Ident, n.pos, n.end, allocator)
 			ident.name = "bool"
 			expr = wrap_pointer(ident, symbol.pointers)
-			return expr, node_to_string(&expr.expr_base), 0, true
+			name := untyped ? fmt.tprintf("u %v", node_to_string(&expr.expr_base)) : node_to_string(&expr.expr_base)
+			return expr, name, 0, true
 		}
 
+		untyped = .Variable in symbol.flags && .Mutable in symbol.flags == false ? true : false
 		expr = symbol_to_expr(symbol, ast_context.fullpath)
 		bl := expr.derived.(^ast.Basic_Lit) or_return
 		bl.tok.text = token_kind_to_str(bl.tok)
 		expr = wrap_pointer(bl, symbol.pointers)
-		return expr, node_to_string(&expr.expr_base), 0, true
+		name := untyped ? fmt.tprintf("u %v", node_to_string(&expr.expr_base)) : node_to_string(&expr.expr_base)
+		return expr, name, 0, true
 
 
 	case SymbolProcedureValue:
--- a/src/server/builtins.odin
+++ b/src/server/builtins.odin
@@ -2,6 +2,7 @@ package server
 
 import "core:fmt"
 import "core:odin/ast"
+import "core:odin/tokenizer"
 import "core:strconv"
 
 // Attempts to resolve the type of the builtin proc by following the rules of the odin type checker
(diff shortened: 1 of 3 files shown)


BradLewis:
Do you have an example of how the constants are handled/ how they should be handled with this? Right now for things like hover with constants I've opted to just show the constant value rather than trying to infer any 'untyped' behaviour.

As an example: 
image

On the symbol itself that information comes from the `flags` and whether or not it contains `.Mutable`.

Nave55:
Some examples to show what I meant.

So these three examples create arrays because they are the same type and all typed or untyped values.

* Untyped literals

``` odin
vals := compress_values(1, 3)
fmt.println(vals)
```

Output -  [1, 3]

* Variables

``` odin
x := 2
y := 7
vals := compress_values(x, y)
fmt.println(vals)
```

Output -  [2, 7]

* Constant and untyped literal

```odin
FOO :: 3
vals := compress_values(FOO, 5)
fmt.println(vals)
```

Output - [3, 5]

These two examples create  a struct because although they are the same type the values differ on whether or not they are typed.

* Variable and untyped literal

```odin
y := 7
vals := compress_values(y, 5)
fmt.println(vals)
````

Output - {v0 = 7, v1 = 5}

* Variable and constant

```odin
FOO :: 3
y := 7
vals := compress_values(FOO, y)
fmt.println(vals)
```

Output -  {v0 = 3, v1 = 7}

```odin
FOO :: 3
y := 7
```

FOO and y look like this in ols by the time compress values receives them

```odin
&Ident{node = Expr{expr_base = Node{pos = Pos{file = "D:/Coding/ols.odin", offset = 100, line = 8, column = 26}, end = Pos{file = "D:/Coding/ols.odin", offset = 103, line = 8, column = 29}, state_flags = Node_State_Flags{}, derived = 0x21510370D48}, derived_expr = 0x21510370D48}, name = "int"}

&Ident{node = Expr{expr_base = Node{pos = Pos{file = "D:/Coding/ols.odin", offset = 105, line = 8, column = 31}, end = Pos{file = "D:/Coding/ols.odin", offset = 106, line = 8, column = 32}, state_flags = Node_State_Flags{}, derived = 0x21510370E10}, derived_expr = 0x21510370E10}, name = "int"}
```
And with this info I don't really see a way of distinguishing between them.

Let me know if this info makes sense. It's possible since I'm not incredibly familiar with the ols codebase that I'm missing something obvious.

BradLewis:
Right I see what you mean. Some information should be in the `symbol.flags` when you call `get_symbol_return_value`. I created a basic example that you highlighted above 

```odin
FOO :: 3

main :: proc() {
	y := 7
	result := compress_values(FOO, y)
}
```

and in this example the `flags` for the 2 args are `bit_set[SymbolFlag]{Variable}` and `bit_set[SymbolFlag]{Variable, Mutable, Local}`. Another part is that the symbol itself will have the underlying `type_expr` if one exists. So in a slightly different version:

```odin
FOO : int : 3

main :: proc() {
	y := 7
	result := compress_values(FOO, y)
}
```

gives 

```
type_expr: &Ident{node = Expr{expr_base = Node{pos = Pos{file = "test/main.odin", offset = 21, line = 2, column = 9}, end = Pos{file = "test/main.odin", offset = 24, line = 2, column = 12}, state_flags = Node_State_Flags{}, derived = 0x102458298}, derived_expr = 0x102458298}, name = "int"}
type_expr: <nil>
```

You can probably use that to help infer what you should be doing in these cases?

Nave55:
I knew about the flags but they are empty in get_symbol_return_value

``` odin
FOO :: 3
y := 7
r: int = 12
vals := compress_values(y, FOO, r)
```

For 'Foo' and 'y' they are both

``` odin
flags = bit_set[SymbolFlag]{}
```

While 'r' isn't empty and looks like

```odin
flags = bit_set[SymbolFlag]{Variable, Mutable, Local}
```

It's because 'r' can be looked up with get_local which 'FOO' and 'y' cannot be.

I'll need to think a bit more for a good solution without having to alter the codebase much.

BradLewis:
I wrote a very simple test based on what you have there:

```odin
@(test)
ast_hover_compress_example :: proc(t: ^testing.T) {
	source := test.Source {
		main = `package test

		main :: proc() {
			FOO :: 3
			y := 7
			r: int = 12
			v{*}als := compress_values(y, FOO, r)
		}
	`,
	}
	test.expect_hover(t, &source, "TBD")
}
```

and adding a very simple log to `get_symbol_return_value`, 

```odin
@(private = "file")
get_symbol_return_value :: proc(
	ast_context: ^AstContext,
	n: ^ast.Expr,
	symbol: Symbol,
	is_proc_return: bool,
	allocator := context.temp_allocator,
) -> (
	expr: ^ast.Expr,
	str: string,
	return_len: int,
	ok: bool,
) {
	log.info(symbol.name, symbol.flags)
...
```

I get the following logs

```
[INFO ] --- [2026-07-23 08:44:12] [builtins.odin:455:get_symbol_return_value()] y bit_set[SymbolFlag]{Variable, Mutable, Local}
[INFO ] --- [2026-07-23 08:44:12] [builtins.odin:455:get_symbol_return_value()] FOO bit_set[SymbolFlag]{Variable, Local}
[INFO ] --- [2026-07-23 08:44:12] [builtins.odin:455:get_symbol_return_value()] int bit_set[SymbolFlag]{Variable, Mutable, Local}
```

Does that not work for you?

Nave55:
For some reason I wasn't seeing what you were. The flags were empty, but I cloned the latest odin build and it works now for some reason. Not sure what changed. 

I made a new commit and added some new tests and it all seems to work properly now. Thanks for the help and let me know if you need me to do anything else.

Capture

BradLewis (approved):
Thanks! All works well for me.

Issue #1554, opened 2026-07-12
[Feature Request] Option to add imports to the bottom of the file


BigBoyBarney:
Hi!

Currently, when letting `ols` autocomplete unimported packages, it puts the import at the top of the file.

Would it be possible to have an option to make `ols` put the imports at the bottom instead?

Cheers!

BradLewis:
Completed in https://github.com/DanielGavin/ols/pull/1571

PR #1571, opened 2026-07-18
Feature: #1554: Add imports to bottom of the file

gxhamster:
When trying to autocomplete a package OLS will place the import after the package declarations. This will introduce a new config option `enable_add_import_to_bottom` that will place the imports at the very bottom of the source file. The option is currently set to `false` as default option.

See feature request [https://github.com/DanielGavin/ols/issues/1554](#1554)

image


BradLewis:
This works for completions, but not for the code action for adding missing imports. Would it be possible to add that as well?

src/server/types.odin
```diff
@@ -438,8 +438,9 @@ OlsConfig :: struct {
 	enable_checker_only_saved:               Maybe(bool),
 	enable_checker_workspace_diagnostics:    Maybe(bool),
 	enable_auto_import:                      Maybe(bool),
+	enable_add_import_to_bottom:             Maybe(bool),
 	enable_code_action_invert_if:            Maybe(bool),
-	struct_fields_underscore_visibility: 	 string,
+	struct_fields_underscore_visibility: 	   string,
```
BradLewis, line 443: Seems some whitespace slipped in here

gxhamster: I have added a new commit that implements the bottom import for the code action as well.

BradLewis (approved):
Thanks!

BigBoyBarney:
Thank you so much for this!

Would it be possible to ignore the last new line when the file otherwise ends with imports?

When saving (and having a final new line added by the editor), the import block is broken up:

https://github.com/user-attachments/assets/a99dad98-a002-4178-9ce5-9d05a57b3a05

gxhamster:
@BradLewis is there any way to have access to the source code (in bytes) to detect things like newlines and other characters that might not neccessarily be parsed within the ast

BradLewis:
> @BradLewis is there any way to have access to the source code (in bytes) to detect things like newlines and other characters that might not neccessarily be parsed within the ast

Yeah the raw file data is added to the `AstContext` under `ast_context.file.src`.

gxhamster:
@BigBoyBarney the import block being broken when the last line is a import has been merged

BigBoyBarney:
@gxhamster Thank you very much, I can confirm it works flawlessly!

PR #1572, opened 2026-07-18
Fold simple package consts in when conditions

### LINKED ISSUE #1562: Support map builtins again after `-bedrock` was added
Looks like map builtin procedures no longer appear in autocomplete, nor you can see the type signature, no go to def etc.

It's because now map procs are in a `when` block enabled by a `MAP_ENABLED` constant, which is derived from `ODIN_BEDROCK`.

Setting `"ODIN_BEDROCK": "false"` in `ols.json` defines does nothing as ols only evaluates direct `when <define>` use, and not a `MAP_ENABLED :: !ODIN_BEDROCK` expression.

```odin
// odin/base/runtime/core_builtin.odin

MAP_ENABLED :: !ODIN_BEDROCK

...

when MAP_ENABLED {
	@builtin
	map_insert :: proc(m: ^$T/map[$K]$V, key: K, value: V, loc := #caller_location) -> (ptr: ^V) {
		key, value := key, value
		return (^V)(__dynamic_map_set_without_hash((^Raw_Map)(m), map_info(T), rawptr(&key), rawptr(&value), loc))
	}
```

It either might be a good opportunity to start adding some basic constant evaluation  mechanism to ols to solve this;
or just hardcode `MAP_ENABLED` for now or something ¯\\_(ツ)_/¯

wishfoundry:
Fixes #1562

This expands WHEN evaluations to check for simple static booleans. 

nothing fancy, just a quality of life improvement

### DIFF
--- a/src/server/ast.odin
+++ b/src/server/ast.odin
@@ -478,6 +478,7 @@ collect_when_stmt :: proc(
 	file: ast.File,
 	file_tags: parser.File_Tags,
 	when_decl: ^ast.When_Stmt,
+	when_expr_map: ^map[string]When_Expr,
 ) {
 	if when_decl.cond == nil {
 		return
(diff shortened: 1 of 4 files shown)


BradLewis (approved):
Thanks!

PR #1575, opened 2026-07-21
Keep range loop headers on one line

Velfi:
## What changed

- keep range-loop headers on one physical line during formatting
- add regression coverage for long collection and range expressions

## Why

`Range_Stmt` currently builds its header with optional line breaks around the loop variables, `in`, and the iterated expression. When a header exceeds `character_width`, those breaks become physical newlines:

```odin
for
	expedition_index
	in
	passage.contract.ship_indices[:passage.contract.ship_count] {
```

Odin's semicolon insertion makes that output invalid. The formatter can therefore turn valid source into code that no longer parses.

This change builds the range header separately and applies `enforce_fit` before attaching the body, preserving valid syntax even when the header exceeds the configured width.

## Validation

- built `odinfmt` from this branch
- added input and snapshot regression cases for long range-loop headers
- formatted a 25-file downstream Odin package that previously reproduced the failure
- verified the formatted downstream package with `odin check`

The full formatter snapshot run reaches an unrelated existing mismatch in `tests/random/demo.odin`: current `master` removes trailing whitespace while its checked-in snapshot still contains it. The focused `for.odin` snapshot passes.

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -1204,25 +1204,29 @@ visit_stmt :: proc(
 			document = cons(document, text("#reverse"), break_with_no_newline())
 		}
 
-		document = cons(document, text("for"))
+		range_document := text("for")
 
 		if v.init != nil {
-			document = cons(document, break_with_space(), visit_stmt(p, v.init), text(";"))
+			range_document = cons(range_document, break_with_space(), visit_stmt(p, v.init), text(";"))
 		}
 
 		if len(v.vals) >= 1 {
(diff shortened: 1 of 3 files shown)


BradLewis:
I'm unable to reproduce this problem on my end without these changes. The snapshots you added seem to just work fine for me. What was the downstream odin file this was having issues with?

Velfi:
Sorry, missed a commit. Can you try the repro again?

LemonatedCat:
I have reported a similar problem in the discord, for me it happened inside switch statement case blocks in particular

BradLewis (approved):
Thanks!

PR #1576, opened 2026-07-22
Feature #1571: Ignore last new line with imports

gxhamster:
When the file ends with a import statement the newline that is usually inserted with each autocompleted import statement should be remove in order to not break up the import block.

Some edge cases still exist if the newline inserted with the import statement during autocomplete is manually removed the next autocompleted import will be on the same line.

Fix for https://github.com/DanielGavin/ols/pull/1571#issuecomment-5048572341

https://github.com/user-attachments/assets/118d51b5-67e6-443d-812d-9c5fe5d76e73


BradLewis (approved):
Thanks!

PR #1581, opened 2026-07-27
Fix crashes due to nil pointers in structs with poly_params

Nave55:
This fixes two issues with poly_params in structs.

* Having more poly_param arguments then parameters would crash due the ```ast.Field``` being nil . 

``` odin
Val :: struct(v: int) {}

main :: proc() {
	val := Val(35, 36){}
}
```
* Having parentheses on a struct with no poly_params would crash due to the ```ast.Field_List``` being nil.

``` odin
Tst :: struct {}

main :: proc() {
	 tst := Tst(){}
}
```

To fix these issues I added some checks that continue if there is a nil pointer.


BradLewis (approved):
Thanks!

PR #1583, opened 2026-07-29
Fixes completions for type switches

Nave55:
Fixes an issue where cases in type switches had missing names.

* Issue
Capture

* Fix was simply to check if symbol.name == "" and if so get its type and add it to used_unions.
Capture


BradLewis (approved):
Thanks!

PR #1584, opened 2026-07-29
Add support for Any type

Nave55:
Adds hover and completion support for the any type.

* Hover
left_id
right_id

* Completion
comple


BradLewis (approved):
Thanks!

PR #1588, opened 2026-07-31
Fix crashes on Maybe with poly params

Nave55:
Did a bit of testing and this seems to prevent the crashing reported on discord.
At least these examples no longer seem to crash.

maybe

Capture

### DIFF
--- a/src/server/generics.odin
+++ b/src/server/generics.odin
@@ -920,9 +920,6 @@ resolve_poly_union :: proc(ast_context: ^AstContext, poly_params: ^ast.Field_Lis
 			continue
 		}
 
-		if param.names == nil {
-			continue
-		}
 		for name in param.names {
 			append(&poly_names, node_to_string(name))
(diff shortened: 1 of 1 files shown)


src/server/generics.odin
```diff
@@ -916,6 +916,13 @@ resolve_poly_union :: proc(ast_context: ^AstContext, poly_params: ^ast.Field_Lis
 	poly_names := make([dynamic]string, 0, context.temp_allocator)
 
 	for param in poly_params.list {
+		if param == nil {
+			continue
+		}
+
+		if param.names == nil {
```
BradLewis: This case is just handled by the for loop no?

Nave55: Yeah you're right don't need the check because it's a slice.

BradLewis (approved):

BradLewis:
Thanks!

PR #1600, opened 2026-08-02
Fix composite literal `return` in single statement case

### LINKED ISSUE #1255: Odinfmt breaks return statement with inline_single_statement_case
If I have a file called `test.odin`:

```odin
package main

parse_if :: proc(state: ^TokenizerState) -> (IfElseStatement, []string) {
    #partial switch _ in state.last_token {
    case: wrong_token_err(state, []string{"`{`"})
    case OpenBraceToken:
        return IfElseStatement{condition, parse_block(state), parse_block(state)}, []string{}
    }
}
```

Then, when I run `odinfmt -path:test.odin`, I get the incorrectly formatted:

```odin
package main

parse_if :: proc(state: ^TokenizerState) -> (IfElseStatement, []string) {
    #partial switch _ in state.last_token {
    case: wrong_token_err(state, []string{"`{`"})
    case OpenBraceToken:
        returnIfElseStatement{condition, parse_block(state), parse_block(state)},
            []string{}
    }
}
```

prnvbn:
fixes https://github.com/DanielGavin/ols/issues/1255

Fix formatting of composite literal returns when a single-statement case wraps.

```odin
case:
   return i8{}
```

now formats as:

```odin
case: return i8{}
```

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -1258,7 +1258,7 @@ visit_stmt :: proc(
 			if is_return_stmt_ending_with_comp_lit_expr(v.results) {
 				document = cons(
 					document,
-					if_break_or_document(empty(), text(" ")),
+					text(" "),
 					visit_exprs(p, v.results, {.Add_Comma}),
 				)
 			} else if !is_return_stmt_ending_with_call_expr(v.results) {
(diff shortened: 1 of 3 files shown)


prnvbn:
@BradLewis lmk your thoughts

BradLewis (approved):
Thanks!

PR #1602, opened 2026-08-02
added themeing support for declarations

Carlyle-Foster:
both the semantic tokens and the TextMate grammar can now distinguish the initial declarations of variable and constants from their uses. i tried to keep backwards compatibility with existing themes as much as possible, so in some cases where `meta.`  scopes were being used incorrectly to refer to concrete tokens i kept them but added correct scopes as well (it's also quite possible that my understanding of TextMate naming conventions is just incomplete)


Carlyle-Foster:
i forgot there's a test suite, the unit tests have already helped me find a few bugs in the semantics tokens code, i should have this wrangled soon enough though

Carlyle-Foster:
`keyword_lut_init()` in `core:odin/tokenizer` at line 85 is crashing nondeterministicly when i run tests, but i think that's an unrelated race condition? there's also some warning from `analysis.odin:1558:internal_resolve_type_expression()`

Carlyle-Foster:
at least the tests are passing

Carlyle-Foster:
> `keyword_lut_init()` in `core:odin/tokenizer` at line 85 is crashing nondeterministicly when i run tests, but i think that's an unrelated race condition? there's also some warning from `analysis.odin:1558:internal_resolve_type_expression()`

the panic in `core:odin/tokenizer` is indeed a [known](https://github.com/odin-lang/Odin/issues/6459) race condition, not my problem i guess

PR #1605, opened 2026-08-03
Distinct types failed to autocomplete

nfginola:
checker: v.type as orig_expr on distinct recursion

Distinct types wrapping a package alias (example: `distinct linalg.Vector4f32`) failed to autocomplete. 
`Selector_Expr` falls through to a default case that resolved the stale outer `Distinct_Type` node instead of the unwrapped type. `internal_resolve_type_expression` has no case for `Distinct_Type` at all, so it just fails silently.

Regression test is added to cover this case.

Noticed this immediately when trying to work with and trying to use any .xyzw 
`Pixel :: distinct linalg.Vector4f32`

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -2242,7 +2242,7 @@ resolve_identifier_expr :: proc(
 
 	#partial switch v in expr.derived {
 	case ^ast.Distinct_Type:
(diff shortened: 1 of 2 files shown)


BradLewis (approved):
Thanks!

PR #1612, opened 2026-08-05
Fix `expand_values` and `compress_values` variadics

BigBoyBarney:
I'm not entirely sure what the intention behind those builtins are, since it's just for `ols` and not valid Odin, so the bug may very well be in the parser.

Currently, whatever is immediately after `compress_values` doesn't show up in the autocomplete, in this case `size_of`.
Adding another random builtin like `snorlax` before it will make `size_of` show up (but notably not `snorlax`).

Changing `...` to `..` fixes this. I have no idea why. 

It's interesting to note that `..` is invalid Odin too, but it seems to work.


BradLewis (approved):
Thanks!

PR #1614, opened 2026-08-06
Do not replace field names with package paths

### LINKED ISSUE #1604: Default param inlay hint shows param name as path if it matches the name of an imported package
```odin
package lib

import "core:math/ease"

foo :: proc (ease := 123) {}
```

```odin
package main

import "./lib"

main :: proc () {
    lib.foo(/*path/to/odin/core/math/ease = 123*/)
}
```

Requires `"enable_inlay_hints_default_params"` naturally.

thetarnav:
fixes #1604

I wasn't able to actually replicate that in tests, but it does fix my issue.


BradLewis (approved):
Thanks!

PR #1615, opened 2026-08-07
Fix enumerated array index completion without dot

thetarnav:
```odin
Foo :: enum {One, Two}
FOO :: [Foo]int{
    // autocomplete here showed "Foo" & "Foo" instead of ".One" & ".Two"
}
```

This fixes this it by reusing the same completions logic for both dot (get_implicit_completion) and without dot (get_comp_lit_completion) completions by reusing `create_enum_completion_item` with added `dot_before_label` param and `append_enum_completion_items` proc to further deduplicate repeated logic.

### DIFF
--- a/src/server/completion.odin
+++ b/src/server/completion.odin
@@ -745,20 +745,12 @@ get_comp_lit_completion :: proc(
 				}
 			}
 			return false
-		case SymbolFixedArrayValue:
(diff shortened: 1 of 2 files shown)


BradLewis (approved):
Thanks!

PR #1616, opened 2026-08-08
Support fake methods for using fields

### LINKED ISSUE #1568: Support `using` in fake methods
```odin
Inner :: struct {foo: int}
Outer :: struct {using inner: Inner}

takes_inner :: proc (inner: Inner) {}

inner: Inner
inner. // takes_inner appears in completions

outer: Outer
outer. // takes_inner doesn't appear in completions
```

thetarnav:
fixes #1568

Looks like `append_method_completion` was missing a recursive call for fields with `using`

### DIFF
--- a/src/server/methods.odin
+++ b/src/server/methods.odin
@@ -1,6 +1,5 @@
 package server
 
-import "core:log"
 import "core:fmt"
 import "core:mem"
 import "core:odin/ast"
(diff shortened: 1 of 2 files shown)


BradLewis (approved):
Thanks!

thetarnav:
@BradLewis just realized this causes a segfault when using fields point to the same struct. Is there an easy way of keying symbols in a map? Perhaps I could just use the `using_type_expr` pointer.

PR #1619, opened 2026-08-09
Update intrinsics

SOBEX:
Updates intrinsics to [9914a57](https://github.com/odin-lang/Odin/blob/9914a57fc7191cfb3b4749909feee66fc01be828/base/intrinsics/intrinsics.odin) and removes where clauses again


BradLewis (approved):
Thanks!

Issue #1620, opened 2026-08-09
Add package namespace references


thetarnav:
```odin
import "core:fmt"

main :: proc () {
    fmt.println("Hello World")
}
```

Triggering go-to-references on any `fmt` symbol should show all fmt package uses in this file.

PR #1624, opened 2026-08-11
Improve references/definition test errors

thetarnav:
Got annoyed with error messages and checking of references in tests while working on #1620 so I've improved them a bit:

- Adds helpers for pretty displaying `common.Location` with code lines
- Removes `expected_excluded` param from `expect_reference_locations`.
  All expected locations are required, any additional are disallowed.
- Fixes couple of bad tests due to these changes.
- Adds similar changes to `expect_definition_locations` and `expect_type_definition_locations`.

image

### DIFF
--- a/src/testing/helpers.odin
+++ b/src/testing/helpers.odin
@@ -74,8 +74,9 @@ source_location_display :: proc (
 
 	max_digits := math.count_digits_of_base(end.line, 10)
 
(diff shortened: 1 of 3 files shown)


BradLewis (approved):
Nice that looks really good now. Very easy to spot what's gone wrong

PR #1630, opened 2026-08-14
Setup spall

thetarnav:
This sets up ols to be easily used with spall for trace profiling.
It won't do anything by default.
Emitting spall traces can be enabled with `SPALL_ENABLED` and full instrumentation with `INSTRUMENTATION_ENABLED` defines.
I've added some traces to most important locations for now—so far it shows how big chunk of most requests is taken by `resolve_entire_file`, even when reporting unused imports is disabled, since each resolve call takes ~400ms (in ols).
This could be modified to support other profilers as well, but spall is just easiest to get into, and its bindings are in core.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 *.exp
 *.sqlite
 *.suo
+*.spall
 *node_modules
 /ols
 /odinfmt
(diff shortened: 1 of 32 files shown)


BradLewis (approved):
Nice! This works for me. It'll be great to have some profiling like this to hunt down those bottlenecks. Just got some merge conflicts now from the other PRs I merged in earlier.

PR #1631, opened 2026-08-16
Add document SymbolAndNode map cache

thetarnav:
After https://github.com/DanielGavin/ols/pull/1629 was merged, the performance should be the same for people with `enable_unused_imports_reporting` enabled—as finding unused imports will resolve entire file and update the cache on `textDocument/didChange` request. But if you had that disabled, ranged-semantic-tokens request wouldn't write to cache so you could re-resolve the same file multiple times, depending on the order of the requests.

This cleans up the symbol map caching a bit.
- Instead of a separate cache map, cached symbol-and-node-maps are now stored directly on the document struct. This is mostly just to reduce code, as we only cache symbols for open documents anyway and the symbols map allocations use document.allocator.
- Removes `resolve_ranged_file` as it's result couldn't be written to cache and the cost of resolving file multiple times—even partially—is greater than the difference between ranged and full resolve. So it's better to just resolve it entirely on first request that needs it and cache so subsequent requests don't need to do it again.
- Makes `resolve_ranged_file(document)` separate from `resolve_entire_file_for_references(document, allocator, flag, target_name)` as `resolve_ranged_file` can be cashed, uses document.allocator and is not configurable, while the other is used only for references and uses special flags to alter or filter the results.

### DIFF
--- a/src/server/references.odin
+++ b/src/server/references.odin
@@ -252,7 +252,7 @@ resolve_references :: proc(
 	}
 
 	target_name := get_target_name(position_context, resolve_flag)
-	symbols_and_nodes := resolve_entire_file(document, resolve_flag, ast_context.allocator, target_name)
+	symbols_and_nodes := resolve_entire_file_for_references(document, ast_context.allocator, resolve_flag, target_name)
 
 	for k, v in symbols_and_nodes {
 		if strings.equal_fold(v.symbol.uri, symbol.uri) && v.symbol.range == symbol.range {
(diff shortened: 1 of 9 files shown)


src/server/references.odin
```diff
@@ -293,17 +293,12 @@ resolve_references :: proc(
 
 
 	arena: runtime.Arena
+	_ = runtime.arena_init(&arena, mem.Megabyte * 40, context.temp_allocator)
```
BradLewis, line 296: This arena is never destroyed anymore. I'm guessing it still should be ?

thetarnav: Not really since it's on a temporary allocator.
I can revert it though as it's not relevant to the PR.

BradLewis: Ah I missed that the allocator has been updated. That's fine like this then.

BradLewis (approved):

PR #1637, opened 2026-08-20
Use nil check instead of 0 comparison for os.Error

karl-zylinski:
Odin compiler crashes on `os.Error == 0` due to an old hack in the compiler. Worked around it by using `nil` check instead. Compiler still needs to be fixed, but probably be removing that `0` comparison capability and making it a real error.

Note: The hack has been removed in Odin compiler now, so there is a real error message instead of the compiler crash. This PR still fixes the compilation error.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


BradLewis (approved):
Thanks!

PR #1639, opened 2026-08-21
Return edit range along with completion text

### LINKED ISSUE #1638: Completion eats the identifier text after the cursor
I'm using ols with zed (odin extension v0.3.16). When using completion, items omit insert/replace ranges, causing zed extension to replace identifier text after the cursor. For example in function params, if I'm prepending a new param, it often removes the following one.

For example, I have `ses|existing` (`|` stands for cursor, which is after `ses` in this case), I auto-complete it to `session` and then `existing` gets replaced with `session|`, but I expect to get `session|existing`.

---

This happens because of this zed behavior: https://github.com/zed-industries/zed/issues/55311

It might be similar to https://github.com/DanielGavin/ols/issues/481

Zelenya:
Resolves #1638

Instead of letting the editor decide/guess the range and operation, we can explicitly return [insert and replace range](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#insertReplaceEdit). This will prevent some editors from unexpectedly replacing the surrounding tokens/identifiers. So, for example, in zed, I can use one with insert with Enter (`ses|existing` → `sessionexisting`) and replace one with `Shift+Enter` (replace `ses|existing` → `session`)

---

I tested this with zed and vs code.

### DIFF
--- a/src/server/completion.odin
+++ b/src/server/completion.odin
@@ -13,6 +13,7 @@ import "core:reflect"
 import "core:slice"
 import "core:strconv"
 import "core:strings"
+import "core:unicode/utf8"
 
 import "src:common"
 import "src:spall"
@@ -181,6 +182,7 @@ get_completion_list :: proc(
 		arg_symbol,
 		config,
 	)
+	add_completion_text_edits(items, &position_context, completion_type, position, config)
 	list.items = items
 	list.isIncomplete = is_incomplete
 	return list, true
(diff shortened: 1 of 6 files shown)


src/testing/testing.odin
```diff
@@ -192,20 +192,7 @@ source_remove_cursor :: proc(src: ^Source) -> (cursor: common.Position) {
...
-		if last == '\r' || ch == '\n' {
-			cursor.line += 1
-			cursor.character = 0
-		} else {
-			cursor.character += 1
-		}
-	}
-
-	return
+	return common.get_relative_token_position(marker_pos, transmute([]u8)source^, 0)
```
Zelenya, line 195: Had to change this to utf-16-aware cursor-search because of this test: https://github.com/DanielGavin/ols/pull/1639/changes#diff-9fede6df4574995e75171e71edb5ba5a63b96859b035f4343ff846cf05e8cd72R52

src/server/completion.odin
```diff
@@ -402,6 +404,72 @@ convert_completion_results :: proc(
...
+
+		new_text := item.insertText.(string) or_else item.label
+		if config.completion_insert_replace_support {
+			item.textEdit = InsertReplaceEdit {
+				newText = new_text,
+				insert  = insert_range,
+				replace = replace_range,
+			}
+		} else {
+			item.textEdit = TextEdit {
```
Zelenya, line 432: Note that this is not backward-compatible. This new fallback sends an explicit prefix-only edit (previously the client could decide to replace the whole token). I think this is better/expected, but let me know if you think otherwise

BradLewis (approved):
Thanks!

PR #1640, opened 2026-08-23
Compile on OpenBSD

JimZAH:
I had a need to use ols on OpenBSD. After making the small adjustment in '/src/common/util.odin' I'm happy to report it works with no noticeable issues so far.


BradLewis (approved):

BradLewis:
Thanks!

PR #1643, opened 2026-08-26
Fix space before "else" in Stroustrup and Allman brace style

### LINKED ISSUE #394: Stroustrup brace style inserting extra space in front of else
The Stroustrup brace style is broken. I believe this used to work correctly. In the latest version of OLS, an extra space is inserted before the `else`:

```odin
    x := 1
    if x == 1 {
        // do somtehing
    }
     else if x == 2 {
        // do something 2
    }
     else {
        // do something else
    }
```

Should instead be:

```odin
    x := 1
    if x == 1 {
        // do somtehing
    }
    else if x == 2 {
        // do something 2
    }
    else {
        // do something else
    }
```

This is my current odinfmt.json:

```json
{
    "$schema": "https://raw.githubusercontent.com/DanielGavin/ols/master/misc/odinfmt.schema.json",
    "character_width": 140,
    "tabs": false,
    "spaces": 4,
    "newline_limit": 4,
    "indent_cases": true,
    "brace_style": "Stroustrup"
}
```

Cristall:
Fixes #394

After the newline there was an additional break document added because of the call to `cons_with_opl`. 
I also fixed that when statements did not respect Stroustrup brace style at all.

### DIFF
--- a/src/odin/printer/visit.odin
+++ b/src/odin/printer/visit.odin
@@ -1033,15 +1033,17 @@ visit_stmt :: proc(
 		set_source_position(p, v.body.end)
 
 		if v.else_stmt != nil {
-			if p.config.brace_style == .Allman ||
-			   p.config.brace_style == .Stroustrup ||
-			   (!p.config.convert_do && block_uses_do(v.body)) {
+			else_on_newline :=
+				p.config.brace_style == .Allman ||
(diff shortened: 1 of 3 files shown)


DanielGavin:
Thanks!

PR #1650, opened 2026-09-01
Implement attribute completion

### LINKED ISSUE #1625: `@(` doesn't complete anything
`get_attribute_completion` in `src/server/completion.odin` has an empty body, and nothing calls it. So typing `@(` gives you nothing back. I confirmed with a completion request inside `@()`: zero items. `#` directives are fine, since `get_directive_completion` is actually implemented and dispatched.

I have a working version locally if you want it. It adds an `attribute_docs` table next to `directive_docs`, names pulled from the compiler's `checker.cpp` so I wasn't guessing at the list. That's 51 of them, leaving out `force` since the compiler just tells you to use `require` instead. Then an `.Attribute` case dispatched the same way `.Directive` already is. Three tests, suite still passes at 891.

One thing I'd want your take on first. I detect the attribute context by scanning backwards through the source for `@` instead of going through the AST. While you're still typing an attribute the declaration under it doesn't parse, so there's no `ast.Attribute` to find, which I'm guessing is also why `DocumentPositionContext` has no field for one. If you'd rather it went through the position context I'm happy to do that, it just needs the parser to hand back something usable for a half-typed attribute.

Worth a PR?

Risu-swift:
Closes #1625.

`get_attribute_completion` was an empty stub with no call site, so typing `@(` returned nothing. This implements it the same way `get_directive_completion` already works: an `attribute_docs` table next to `directive_docs`, an `.Attribute` completion type, and dispatch in `get_completion_list`.

The attribute names come from the compiler's `src/checker.cpp` rather than being written by hand, 51 of them. `force` is left out because the compiler rejects it with "'force' was replaced with 'require'".

**On the detection method.** `is_in_attribute` scans the source backwards for `@` instead of going through the AST. An attribute is typed before the declaration it applies to, so while you are still writing one there is no parseable declaration to hang an `ast.Attribute` off, which is also why `DocumentPositionContext` has no field for one. The scan:

- stops at `)`, so a finished `@(private)` earlier on the line does not keep offering attributes
- stops at `=`, so `@(deferred_none=my_proc` completes identifiers rather than attribute names, unless a comma has already closed that pair (`@(link_name="x", requ`)
- ignores an `@` inside a comment, checked against `file.comments`

Happy to move this to the position context instead if you would rather, it just needs the parser to hand back something usable for a half-typed attribute.

**Known gaps.** The parenthesis-less form (`@private`) is not offered, ols returns early on that shape before completion type selection. An attribute list split across lines is not completed past the first line, since the scan stops at a newline.

**Tests.** Seven tests in `tests/completions_test.odin`, covering the plain and partial cases and each of the three ways the scan bails out. Suite goes 894 to 901, all passing. The 9 failures on my machine are identical on a clean `master` checkout, all hover/doc-comment related and unrelated to this change.

### DIFF
--- a/src/server/completion.odin
+++ b/src/server/completion.odin
@@ -118,10 +118,6 @@ get_completion_list :: proc(
 		completion_type = .Directive
 	}
 
-	// Attributes are matched against the source rather than the AST. A
-	// half-typed `@(` leaves the declaration it belongs to unparseable, so by
-	// the time the user wants this completion there is no Attribute node to
-	// find — the same reason the position context has no field for one.
 	if is_in_attribute(&position_context) {
 		completion_type = .Attribute
 	}
@@ -747,17 +743,6 @@ get_completion_description :: proc(ast_context: ^AstContext, symbol: Symbol) -> 
 	return get_short_signature(ast_context, symbol)
 }
 
-/*
-	Is the position inside an attribute, i.e. after an `@` or inside its `(...)`?
-
-	Scans the source backwards from the cursor instead of consulting the AST: an
-	attribute is typed before the declaration it applies to, so while the user is
-	writing one there is no complete declaration to hang an ast.Attribute off.
-
-	Stops at a newline, since an attribute's `@` is always on the same line as
-	the name being completed, and bails on `)` so a finished `@(private)` on the
-	same line does not keep offering attributes afterwards.
-*/
 is_in_attribute :: proc(position_context: ^DocumentPositionContext) -> bool {
 	src := position_context.file.src
 	pos := position_context.position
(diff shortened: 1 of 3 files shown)


src/server/completion.odin
```diff
@@ -117,6 +118,14 @@ get_completion_list :: proc(
 		completion_type = .Directive
 	}
 
+	// Attributes are matched against the source rather than the AST. A
+	// half-typed `@(` leaves the declaration it belongs to unparseable, so by
+	// the time the user wants this completion there is no Attribute node to
+	// find — the same reason the position context has no field for one.
+	if is_in_attribute(&position_context) {
```
BradLewis, line 121: Can you remove these large comments?

Risu-swift: Done, removed them.

Risu-swift:
The CI failure is in `src/odin/printer/visit.odin:1467`: `'Inline_Asm_Expr' is not declared by 'ast'`. That file is not touched in this PR.

odin-lang/Odin@8181e35 (Aug 31) replaced `ast.Inline_Asm_Expr` with the new `Asm_Template`/`Asm_Spec`/`Asm_Register`/`Asm_Clobber` nodes, and CI builds against `ODIN_BRANCH: master`.

BradLewis (approved):
Thanks!

PR #1653, opened 2026-09-01
Temporary fix for asm so that ols can be built

thetarnav:


BradLewis (approved):

BradLewis:
Thanks I was meaning to do this today but didn't end up getting the time to.

PR #1657, opened 2026-09-02
Feature: optionally preserve blank lines between struct fields

giuinktse7:
## Motivation

`odinfmt` currently collapses blank lines between struct fields, but I often want to separate fields visually into logical groups.

For example, this:

```odin
Settings :: struct {
	window_width:  int,
	window_height: int,
	target_fps:    int,

	fullscreen:    bool,
	vertical_sync: bool,
}
```

becomes:

```odin
Settings :: struct {
	window_width:  int,
	window_height: int,
	target_fps:    int,
	fullscreen:    bool,
	vertical_sync: bool,
}
```

It is possible to preserve the visual separation by adding a comment, but a comment should not be necessary solely for formatting:

```odin
Settings :: struct {
	window_width:  int,
	window_height: int,
	target_fps:    int,

	// Display options
	fullscreen:    bool,
	vertical_sync: bool,
}
```

## Proposed Solution

This PR adds an opt-in `preserve_struct_blank_lines` setting:

```json
{
	"preserve_struct_blank_lines": true
}
```

When enabled, the formatter preserves blank lines between struct fields while respecting the existing `newline_limit`
setting.

Each blank-line-separated section is also aligned independently, similar to `gofmt`. A long field name in one logical
section therefore does not add padding to fields in another section. This applies to both `align_struct_fields` and
`align_struct_declarations`.

Comments attached to fields are treated as part of those fields when identifying section boundaries. Blank lines inside multiline block comments do not create sections, while an actual blank line between a field and the next field or its documentation does.

When `newline_limit` is zero, blank lines are removed and alignment continues across the whole struct because no visible section boundary remains.

The option defaults to false, so existing formatting behavior remains unchanged.

### Example

Fields previously aligned across the entire struct:

```odin
Fields :: struct {
      short:            int,
      medium_name:      int,

      much_longer_name: int,
      x:                int,
}
```

With `preserve_struct_blank_lines` enabled, each blank-line-separated section now has its own alignment width:

```odin
Fields :: struct {
      short:       int,
      medium_name: int,

      much_longer_name: int,
      x:                int,
}
```

## Alignment Improvements

This PR also fixes an inconsistency where `#subtype` was included in declaration alignment widths but omitted from field-type alignment widths. For example, with `align_struct_fields` enabled:

```odin
  Fields :: struct {
        #subtype embedded: Embedded,
        regular: int,
  }
```

Previously, `#subtype` was omitted when calculating the maximum width, producing:

```odin
  Fields :: struct {
        #subtype embedded: Embedded,
        regular:  int,
  }
```

After including `#subtype` in the shared width calculation:

```odin
  Fields :: struct {
        #subtype embedded: Embedded,
        regular:           int,
  }
```

## Verification

Added snapshot coverage for:

- Preserving blank lines between field groups
- Aligning each field group independently
- Section alignment with both `align_struct_fields` and `align_struct_declarations`
- Leading, trailing, line, and multiline block comments around section boundaries
- Ignoring blank lines contained within multiline block comments
- Accounting for `#subtype` prefixes during alignment
- Limiting excessive blank lines according to `newline_limit`
- Using whole-struct alignment when `newline_limit` is zero
- Retaining the existing collapsing behavior when the option is disabled

The `odinfmt` snapshot suite passes.

### DIFF
--- a/tools/odinfmt/tests/preserve_struct_field_blank_lines/.snapshots/structs.odin
+++ b/tools/odinfmt/tests/preserve_struct_field_blank_lines/.snapshots/structs.odin
@@ -1,0 +1,27 @@
+package preserve_struct_field_blank_lines
+
+Settings :: struct {
+	window_width:  int,
+	window_height: int,
+	target_fps:    int,
+
+	fullscreen:    bool,
+	vertical_sync: bool,
+}
+
+With_Comments :: struct {
+	first:  int, // trailing comment
+
+	// The second group.
+	second: int,
+
+	// The third group follows this comment without a blank line.
+	third:  int,
+}
+
+Capped_By_Newline_Limit :: struct {
+	first:  int,
+
+
+	second: int,
+}
--- a/tools/odinfmt/tests/preserve_struct_field_blank_lines/.snapshots/structs.odin
+++ b/tools/odinfmt/tests/preserve_struct_field_blank_lines/.snapshots/structs.odin
@@ -9,18 +9,62 @@ Settings :: struct {
 	vertical_sync: bool,
 }
 
+Different_Section_Widths :: struct {
+	short:       int,
+	medium_name: int,
+
+	much_longer_name: int,
+	x:                int,
+}
+
+Comment_Within_Section :: struct {
+	short:            int,
+	// This comment does not start a new section.
+	much_longer_name: int,
+}
+
+Subtype_Alignment :: struct {
+	#subtype embedded: Embedded,
+	regular:           int,
+}
+
+Block_Comment_Within_Section :: struct {
+	short:            int,
+	/* This comment contains an empty line.
+
+	It does not create a new field section. */
+	much_longer_name: int,
+}
+
+Trailing_Block_Comment_Within_Section :: struct {
+	short:            int,
+	/* This trailing comment contains an empty line.
+
+	It does not create a new field section. */
+	much_longer_name: int,
+}
+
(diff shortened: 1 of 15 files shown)


tools/odinfmt/tests/preserve_struct_field_blank_lines/.snapshots/structs.odin
```diff
@@ -0,0 +1,27 @@
+package preserve_struct_field_blank_lines
+
+Settings :: struct {
+	window_width:  int,
+	window_height: int,
+	target_fps:    int,
```
BradLewis, line 6: I wonder with this if we should align each section separately? Similarly to how `gofmt` handles it.

giuinktse7: Ah, I agree! That behavior is better.

I've updated the PR (& PR body) to use the `gofmt` style, see:
- new changes: [157cf11..1580cff](https://github.com/DanielGavin/ols/pull/1657/changes/157cf11e0d4f8d3703892e19c6c06342f69fa076..1580cffb9f606859a4d685e65fa5f132893f8537#diff-85fbff4ed57c444bece8893586270db7e44a71edc3d4a13993e0e8e6c15607d3)
- The main functionality change: [feat: use section-scoped struct field alignment across blank lines](https://github.com/DanielGavin/ols/pull/1657/changes/dab8f3893db5e8b75e00adb12f6e0484eb5c30ca)

PR #1663, opened 2026-09-05
Improve proc group scoring to match odin

thetarnav:
There was a recent PR to Odin improving scoring of procedures when resolving procedure groups: https://github.com/odin-lang/Odin/pull/7227
Initially I wanted just to include the same changes, but it looks like OLS was missing few additional scoring rules as well.

This PR
- adds a new test file with tests directly taken from Odin's internal tests (https://github.com/odin-lang/Odin/blob/master/tests/internal/test_proc_group_type_inference.odin) — I think that a dedicated test file just for syncing proc group scoring between odin and ols is a good idea, instead of just adding everything to existing hover tests or whatever.
- improves procedure group scoring rules in `analysis.odin` to match odin

### DIFF
--- a/src/server/analysis.odin
+++ b/src/server/analysis.odin
@@ -794,6 +794,7 @@ CallArg :: struct {
 	is_nil:            bool,
 	bad_expr:          bool,
 	has_symbol:        bool,
+	is_constant:       bool,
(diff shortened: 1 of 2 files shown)


BradLewis (approved):
Thanks!

PR #1689, opened 2026-09-17
Skip hidden directories during package discovery

harold-b:


BradLewis (approved):
Thanks for all of these!

harold-b:
Thank **you**!

PR #1698, opened 2026-09-21
Add various Objective-C enhancements

harold-b:
- Objective-C methods returning `instancetype` now resolve to the receiver’s concrete class. This works through local variables, inherited methods, chained calls, and parentheses, so completion, hover, and navigation agree on the type.

- Definition links point to the objc class' method or ivar actual source file, including inherited members. Completion also works through selector chains such as `self.impl->`.
- ObjC ivars appear before methods in completion. Parent ivars are included, the nearest class wins when there are name collisions.
- ObjC methods without the `@objc_name` attribute are implicitly indexed via the same rule Odin uses. It removes the matching class name's prefix from their Odin procedure name. Explicit names still take precedence.
- Added support for `Objc_Block` capture completion using the handler’s parameter types. 
- Added hover formatting for `Objc_Block` that separates callback parameters from captures, showing captures in brackets (`actual_param, [captured_param1, captured_param2]`).
- Added an `Objc_Block` code action that generates a handler `proc` using the expected callback signature and captured values. It handles unfinished calls and missing commas, keeps known names and types, and leaves unknown types blank.

- Member completion in an Objective-C-heavy file improved from 1,242.75 ms to 23.50 ms median, nearly 53× faster. The optimization avoids resolving the same receiver type for each fake method. (This is not an objc-specific improvement, but any fake method indexing, though this particular file had a very large improvement, along with an objc-specific improvement when doing ivars)


harold-b:
Fixing CI-surfaced issues...

harold-b:
✅

BradLewis (approved):
Thanks! I don't have a good objc setup to really test this out, but overall it's working for me and it looks good so LGTM!

